PR #4, opened 2026-08-25
Fix CI: coverage reporting, e2e-mock merges, visual & skill checks

### LINKED ISSUE #5: Visual regression: profile & settings modal screenshot mismatch
### Area

Web app (web/app)

### What happened

The visual E2E test `profile & settings modal, profile tab open` in `e2e/tests/visual/profile-settings.visual.spec.ts` is failing on main. The test compares the rendered profile settings modal screenshot against a committed baseline PNG, and 309 pixels (0.01% of all image pixels) differ from the baseline. The test fails consistently on both the initial run and retry #1.

Test output:
```
Error: expect(page).toHaveScreenshot(expected) failed
309 pixels (ratio 0.01 of all image pixels) are different.
```

Generated diff images are available in the test-results directory.

### What you expected

The rendered profile settings modal screenshot should match the committed baseline PNG without differences.

### Steps to reproduce

1. Run the E2E visual test suite on main: `cd web/app && npx playwright test --config e2e/playwright.config.mock-llm.visual.ts`
2. The test `profile & settings modal, profile tab open` (line 4 in `e2e/tests/visual/profile-settings.visual.spec.ts`) fails
3. View the diff at `test-results/visual-profile-settings.vi-*-ings-modal-profile-tab-open-chromium-desktop/profile-settings-modal-diff.png`

### Version or commit

e638cf76f3eb77209f111ff3e9a70b990dc5bbc3

### Environment

CI environment: E2E (mock backend) workflow running on ubuntu-latest with Playwright in the pgvector:pg16 container. Browser: chromium-desktop.

### Additional context

The test already has proper masking in place for dynamic content (email, initials, "Member since" date). The 309-pixel difference suggests a UI/CSS change affecting the modal layout, identity card, or email input styling. This could be:

- CSS/styling changes to the profile modal or identity card component
- Font rendering or layout differences
- Changes to component positioning or spacing
- Modifications to the email input field appearance

Related test file: `web/app/e2e/tests/visual/profile-settings.visual.spec.ts`
Related fixture: Profile settings modal fixture

### LINKED ISSUE #6: Fix: .claude/skills symlinks are broken on main
### Area

CI / Docker / build

### What happened

The `Skills Symlink Check` workflow on main is failing. `make check-skill-symlinks` reports nine `.claude/skills/` entries that exist as plain files (containing the intended symlink target as text) instead of actual symlinks:

```
❌ .claude/skills/address-pr-feedback exists but is not a symlink (should link into .agents/skills/address-pr-feedback)
❌ .claude/skills/gh-issue exists but is not a symlink (should link into .agents/skills/gh-issue)
❌ .claude/skills/gh-pr exists but is not a symlink (should link into .agents/skills/gh-pr)
❌ .claude/skills/gh-run-watch exists but is not a symlink (should link into .agents/skills/gh-run-watch)
❌ .claude/skills/playwright-cli exists but is not a symlink (should link into .agents/skills/playwright-cli)
❌ .claude/skills/playwright-component-testing exists but is not a symlink (should link into .agents/skills/playwright-component-testing)
❌ .claude/skills/playwright-e2e exists but is not a symlink (should link into .agents/skills/playwright-e2e)
❌ .claude/skills/playwright-trace exists but is not a symlink (should link into .agents/skills/playwright-trace)
❌ .claude/skills/playwright-version exists but is not a symlink (should link into .agents/skills/playwright-version)
make: *** [Makefile:399: check-skill-symlinks] Error 1
```

Job 97989098401 (2026-08-25T22:20:41Z) only surfaced the four `playwright-*` entries in its truncated log, but all nine listed above were affected — `make check-skill-symlinks` still fails after fixing only the four originally reported.

### What you expected

Every `.claude/skills/<name>` should be a relative symlink into `.agents/skills/<name>`, per the source-of-truth pattern documented in AGENTS.md.

### Steps to reproduce

1. On main branch, check the `.claude/skills/` directory
2. Run `make check-skill-symlinks`
3. The check fails because nine directories exist as plain files instead of symlinks

### Version or commit

ec4dea4f0745c84df68e3edffb15abb8562883c2

### Environment

CI environment (ubuntu-latest), GitHub Actions

### Additional context

According to the workflow documentation in `.github/workflows/skills-check.yml`, AGENTS.md documents `.agents/skills/` as the vendor-neutral source of truth, with `.claude/skills/` holding relative symlinks into it, and CLAUDE.md / GEMINI.md as symlinks to AGENTS.md.

Fix (all nine entries):

```bash
cd .claude/skills
for name in address-pr-feedback gh-issue gh-pr gh-run-watch \
  playwright-cli playwright-component-testing playwright-e2e \
  playwright-trace playwright-version; do
  rm -f "$name"
  ln -s "../../.agents/skills/$name" "$name"
done
```

### LINKED ISSUE #7: coverage-summary reports a false total when go tool cover fails
When `go tool cover -func` exits with a non-zero status partway through processing, the `coverage-summary` make target discards the exit code and reports a fabricated total instead of failing.

## Details

The old recipe piped `go tool cover -func` through `tail -1 | awk '{print $NF}'`, which discards the process's exit code entirely.

`go tool cover -func` streams one coverage line per function, then can exit non-zero (exit code 2) if the profile references a source file that isn't present — for example, generated Ent code that was never regenerated with `go generate ./ent`.

Because the exit code was discarded and `tail -1` just grabs whatever the last printed line happened to be, the recipe would print that last *function's* individual coverage percentage and label it as the overall total — with no indication anything had failed.

### Impact

In practice, this caused CI to report ~94.6% coverage for the `go-e2e` and `go-api` flags, when the real numbers were in the low-to-mid 30s. The 94.6% was actually just `func main()`'s own coverage, printed right before `cover -func` died on a missing generated Ent file.

### Reproduction

```
$ go tool cover -func=prof.out; echo "exit=$?"
example.com/covtest/p/a.go:3:   A   100.0%
exit=2
cover: open /tmp/covtest/p/gone.go: no such file or directory

$ go tool cover -func=prof.out 2>/dev/null | tail -1 | awk '{print $NF}'
100.0%      # <- wrongly reported as the total
```

### Solution

Capture both the command's output and its exit status, and require the captured line to actually start with `total:` before treating it as the real total. Otherwise, the command now fails loudly instead of reporting a fabricated number.

unlikelyzero:
Fixes #7
Fixes #5
Fixes #6

Four independent CI checks were red on `main`: the coverage summary
tooling, the `merge` job in `e2e-mock.yml`, a visual regression test,
and the skills symlink check. None of the fixes depend on each other,
but they're bundled here because each was found while chasing down
why coverage reporting looked wrong — see the commits for a clean
split. This PR gets CI back to green across all four.

## Summary
- `coverage-summary` piped `go tool cover -func` through `tail -1`,
  which discarded the exit code and printed the last per-function
  line as the total whenever cover died partway through (e.g. a
  missing source file). It now checks the exit status and requires
  the captured line to start with `total:`, failing loudly otherwise.
- The `merge` job in `e2e-mock.yml` never ran `go generate ./ent`
  before merging Go coverage profiles, so `cover -func` died on the
  first generated Ent file it couldn't find — this is why `go-e2e`
  and `go-api` were reporting ~94.6% (the coverage of `func main()`)
  instead of their real totals in the low-to-mid 30s.
- The same merge step hardcoded `gocov-1 gocov-2 gocov-visual` while
  the matrix runs `SHARD_TOTAL` shards, silently dropping shards 3
  and 4. The artifact list is now derived from `SHARD_TOTAL`, mirroring
  the frontend merge step that already does this correctly.
- The profile settings modal's committed visual baselines (desktop and
  mobile) still had the old, longer subtitle copy, which wraps
  differently than the current text and cascades into every element
  below it. Regenerated both baselines against the CI e2e Docker image
  so they match the current UI.
- Nine `.claude/skills/` entries had been materialized as plain text
  files (containing their symlink target as content) instead of real
  symlinks, breaking `check-skill-symlinks`. Restored all nine as
  relative symlinks into `.agents/skills/`.

## Test plan
- [x] `make coverage-summary PROFILE=coverage/go-unit.out` prints
      `coverage/go-unit.out: 33.4%`
- [x] A malformed profile referencing a missing source file now exits
      non-zero and prints no percentage (previously exited 0 with a
      false percentage)
- [x] Workflow YAML validated for syntax
- [x] `profile & settings modal, profile tab open` visual test passes
      (desktop and mobile) against the CI e2e Docker image
- [x] `make check-skill-symlinks` passes
- [ ] CI on this PR: confirm the `merge` job's `go-e2e`/`go-api`
      coverage-summary lines land in the low-to-mid 30s rather than 94.6%
- [ ] CI on this PR: all four previously-red checks pass


GoriVix (approved):
LGTM

Thanks I was trying to fix the skills thing and it was being annoying

PR #9, opened 2026-08-25
Warn on AUTO_MIGRATE, support arm64 for visual-docker

### LINKED ISSUE #8: Local dev bootstrap: AUTO_MIGRATE check and arm64 visual-docker support
### Area

Local dev tooling (scripts/, e2e/scripts/)

### What happened

Bootstrapping a fresh local stack for the mock-LLM visual regression suite surfaced two rough edges:

1. **`scripts/check-env.sh` doesn't validate `AUTO_MIGRATE`.** It already checks `.env` presence, `JWT_SECRET`/`JWT_REFRESH_SECRET`/`TOKEN_ENCRYPTION_SECRET` quality, and Docker/tool availability — but a fresh `.env` ships with `AUTO_MIGRATE=false`, and `make run-mock`/`make dev-up` against a freshly created (unmigrated) database fails with a fatal `relation "roles" does not exist` panic instead of a clear preflight message. Someone bootstrapping for the first time hits this only after `check-env` already reported everything green.
2. **`e2e/scripts/visual-docker.sh` doesn't run on Apple Silicon.** The script resolves the CI image tag and checks `docker manifest inspect` to decide whether to build locally; that check passes because the tag *does* exist in GHCR, but only as an `amd64` manifest. `docker run` then fails with `no matching manifest for linux/arm64/v8 in the manifest list entries: no match for platform in manifest: not found`, because the script never passes a `--platform` flag. Working around it means manually re-running the same `docker run` invocation from the script with `--platform linux/amd64` added by hand.

### What you expected

- `make check-env` should flag `AUTO_MIGRATE=false` against a database that hasn't been migrated yet (or at minimum print guidance that a fresh DB needs `AUTO_MIGRATE=true` for the first `run-mock`/`dev-up`).
- `e2e/scripts/visual-docker.sh` should detect a non-amd64 host (`uname -m`) and pass `--platform linux/amd64` automatically, since the published CI image is amd64-only and baselines must render identically to CI regardless of host architecture.

### Steps to reproduce

1. `cp .env.example .env`, fill in `JWT_SECRET`/`JWT_REFRESH_SECRET`/`TOKEN_ENCRYPTION_SECRET`, leave `AUTO_MIGRATE=false` (the shipped default).
2. `make check-env` — reports clean.
3. `make db-up && make run-mock` — fails with `pq: relation "roles" does not exist`.
4. On Apple Silicon: `npm run e2e:mock-llm:visual:docker` (or `:update`) — fails with a `no matching manifest for linux/arm64/v8` Docker error instead of running the suite.

### Environment

macOS (Apple Silicon), Docker Desktop, local dev bootstrap for the mock-backend E2E/visual suite.

unlikelyzero:
Fixes #8

## Summary
- `scripts/check-env.sh` now warns when `AUTO_MIGRATE` isn't `"true"`.
  It can't tell whether a given database is already migrated, so a
  fresh one (right after `make db-up`) previously only surfaced this
  as a fatal `relation "roles" does not exist` panic from the API
  server on first boot, well past the point `check-env` said
  everything was fine.
- `e2e/scripts/visual-docker.sh` now detects a non-amd64 host
  (`uname -m`) and adds `--platform linux/amd64` to both the local
  `docker build` fallback and the `docker run` invocation. The
  published CI image is amd64-only; without this, the script fails
  outright on Apple Silicon with `no matching manifest for
  linux/arm64/v8`, and a local arm64 build (had one been produced)
  would render fonts differently than CI, reintroducing the very
  baseline-diff-that-looks-like-a-UI-regression problem this script
  exists to avoid.

## Test plan
- [x] `make check-env` with `AUTO_MIGRATE=false` prints the new
      warning; with `AUTO_MIGRATE=true` it's silent on that line.
      Exit code stays 0 either way (advisory, not blocking).
- [x] `make check-env` still reports "Environment looks good" overall.
- [x] Ran `e2e/scripts/visual-docker.sh` end-to-end on Apple Silicon
      against a local mock-LLM API + Postgres: prints "Host arch is
      arm64 — using --platform linux/amd64 so rendering matches CI",
      pulls/runs the amd64 image successfully (no manifest error),
      and the suite executes (10 passed, 2 failed — the 2 failures
      are the pre-existing stale visual baseline this branch doesn't
      include the fix for yet, not a regression from this change).

### DIFF
--- a/scripts/check-env.sh
+++ b/scripts/check-env.sh
@@ -97,6 +97,18 @@ if [ -n "$db_port" ] && [ "$db_port" != "$compose_port" ]; then
   warn "DB_PORT ($db_port) != compose exposed port CHATAPP_DB_EXPOSED_PORT ($compose_port)"
 fi
 
+# --- migrations ---
+# This script never connects to the database, so it cannot tell whether the
+# schema is already migrated — AUTO_MIGRATE=false is entirely normal against
+# an existing DB. It is only a trap against a *fresh* one: cmd/api-server
+# skips migrations and then dies on boot with `relation "..." does not
+# exist` instead of a clear message, which is what a first `make run-mock`
+# against a just-created `make db-up` volume hits. Warn, don't fail — this
+# script has no way to distinguish the two cases.
+if [ "$(envval AUTO_MIGRATE)" != "true" ]; then
+  warn "AUTO_MIGRATE is not \"true\" — fine against an already-migrated DB, but the first run against a fresh one (e.g. right after 'make db-up') will fail to boot with a 'relation ... does not exist' error. Set AUTO_MIGRATE=true in .env for that first run."
+fi
+
 # --- secrets ---
 # Mirrors internal/auth.ValidateSecret and datastore.ValidateTokenEncryptionSecret
 # exactly: the server hard-fails at boot on any of these conditions, so this
--- a/web/app/e2e/scripts/visual-docker.sh
+++ b/web/app/e2e/scripts/visual-docker.sh
@@ -44,6 +44,23 @@ set -euo pipefail
 repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd)"
 app_dir="web/app"
 
+# The published image (ci-image.yml) is amd64-only. On a non-amd64 host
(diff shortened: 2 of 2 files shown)


GoriVix (approved):
LGTM, merging

PR #10, opened 2026-08-25
Add unit tests for pure-logic internal packages

### LINKED ISSUE #11: NormalizeAndValidateChatTags does not honor its documented nil-in/nil-out contract
## Summary

`NormalizeAndValidateChatTags` in `internal/modeltypes/chat_tags.go` has a documented contract in its doc comment: "nil input returns nil output (useful for PATCH-style 'unset' handling)." However, the implementation does not honor this contract.

## The Bug

The function performs:
```go
normalized := make([]string, len(tags))
// ... validation and processing ...
return normalized, nil
```

When `tags` is `nil`, `len(tags)` is 0, so `make([]string, 0)` creates a non-nil, zero-length slice. This non-nil empty slice is then returned, not `nil`.

## Impact

Callers rely on the documented contract to distinguish two semantically different cases:
- **"field omitted" (PATCH unset):** `nil` return value
- **"explicitly cleared to empty":** non-nil empty slice return value

Because the function always returns a non-nil slice for any input (including `nil`), callers cannot reliably detect the "unset" signal the function is documented to provide.

## Fix

This issue is fixed in PR #10. The fix adds an explicit guard at the top of the function:
```go
if tags == nil {
    return nil, nil
}
```

This guard is placed before any length checks or processing, ensuring the documented nil-in/nil-out contract is actually honored.

unlikelyzero:
Fixes #11

Stacked on #4 — checkpoint 2 of the coverage handoff plan.

## Summary
- Table-driven unit tests for `internal/buildinfo`, `internal/apicontext`,
  `internal/modeltypes`, `internal/featuregate`, `internal/logging`,
  `internal/metering`, and `internal/plugins` — the pure-logic packages
  with no DB, fixtures, or network dependency, matching the style in
  `internal/agent/tools/tool_helpers_test.go`.
- `internal/userhooks` and `internal/providers` are intentionally left
  without test files: both are declarations only (a `Hook`/`Meter` type,
  package-level function vars, and interfaces respectively) with no
  executable statements to cover.
- Writing the `modeltypes` tests surfaced a real bug:
  `NormalizeAndValidateChatTags`'s doc comment promises nil input
  produces nil output (documented for PATCH-style unset handling), but
  `make([]string, len(tags))` on a nil slice returns a non-nil empty
  slice, so the documented contract never actually held. Fixed with an
  explicit nil check.

## Coverage note
Every function touched by these tests is now at 100% (one branch in
`NewLogger` sits at 95.7% — the `zap.Config.Build()` internal-error
path, not practically triggerable without mocking zap internals).
However, the overall `go-unit` total only moved from 33.4% to 33.6% —
these packages are individually tiny (a few dozen statements total)
against a large `-coverpkg` denominator, so they can't move a
codebase-wide percentage much regardless of how thoroughly they're
tested. The handoff doc's ≥40% target for this checkpoint looks like
an optimistic estimate rather than something achievable from this
specific package list; the real per-checkpoint movement should come
from checkpoint 3 onward (provider adapters, datastore, agent
internals), which have far more uncovered surface.

## Test plan
- [x] `go test ./internal/buildinfo/... ./internal/apicontext/...
      ./internal/modeltypes/... ./internal/featuregate/...
      ./internal/logging/... ./internal/metering/... ./internal/plugins/...`
      passes
- [x] `make test-ci` passes (full suite green)
- [x] `make coverage-summary PROFILE=coverage/go-unit.out` →
      `33.6%` (from 33.4%)
- [x] `gofmt`/`go vet` clean

### DIFF
--- a/internal/apicontext/user_test.go
+++ b/internal/apicontext/user_test.go
@@ -1,0 +1,62 @@
+package apicontext
+
+import (
+	"context"
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestWithUserIDAndUserIDFrom(t *testing.T) {
+	tests := []struct {
+		name   string
+		seed   func(ctx context.Context) context.Context
+		wantID uuid.UUID
+		wantOK bool
+	}{
+		{
+			name: "context carries the user id it was given",
+			seed: func(ctx context.Context) context.Context {
+				return WithUserID(ctx, uuid.MustParse("11111111-1111-1111-1111-111111111111"))
+			},
+			wantID: uuid.MustParse("11111111-1111-1111-1111-111111111111"),
+			wantOK: true,
+		},
+		{
+			name: "bare context has no user id",
+			seed: func(ctx context.Context) context.Context {
+				return ctx
+			},
+			wantID: uuid.Nil,
+			wantOK: false,
+		},
+		{
+			name: "nil uuid is still a present value",
+			seed: func(ctx context.Context) context.Context {
+				return WithUserID(ctx, uuid.Nil)
+			},
+			wantID: uuid.Nil,
+			wantOK: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := tt.seed(context.Background())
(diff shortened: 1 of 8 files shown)


GoriVix (approved):
LGTM, merging

PR #12, opened 2026-08-25
Add shared unit test suite for OpenAI-compatible provider adapters

unlikelyzero:
Stacked on #10 (checkpoint 2), which stacks on #4 (checkpoint 1) —
checkpoint 3 of the coverage handoff plan, plus a follow-up closing
the claude/openai/gemini gap noted below.

## Summary
Mistral, DeepSeek, Qwen, Xiaomi, and Local (`internal/agent/provider/
{mistral,deepseek,qwen,xiaomi}_adapter.go` and `local.go`) are
near-identical wrappers over the OpenAI Chat Completions API — same
shape, different constants and error text. One shared table-driven
suite (`openai_compatible_adapters_test.go`) drives all five through
`httptest` servers rather than duplicating the same assertions five
times:

- text responses and tool-call rounds
- `AppendToolResults` folding into the next request (verified by
  inspecting the actual outgoing request body, not adapter-private
  fields)
- `ForceFinalResponse` stripping the tool list and appending the nudge
  message
- streaming via a text-delta handler
- a plain API error (400, non-safety-violation body) passing through
  wrapped but *not* misclassified as a safety violation
- disabled-tool filtering at adapter construction

Local gets two dedicated tests beyond the shared table: it has no
streaming path at all (no `CallStreaming`, and `Call` ignores any
delta handler even when one is set), which the shared table can't
express.

Every function in these five files now sits at 83–100% coverage (was
0–89%, with `Call`/`CallStreaming`/`AppendToolResults`/
`ForceFinalResponse`/`call`/`toGenerateResponse` all previously
untested). No test hits real network — everything is a local
`httptest.Server` with `option.WithBaseURL`, matching the pattern
already established in `chat_completions_stream_test.go`.

## Follow-up: claude/openai/gemini Call/ForceFinalResponse paths
Claude, OpenAI, and Gemini already had substantial dedicated test
files, but each adapter's core `Call`/`ForceFinalResponse`/streaming
paths were completely untested since they require a live HTTP
round-trip through their respective SDK clients — the earlier PR
description below flagged this as its own follow-up rather than
bolting mismatched tests onto the shared-adapter suite.

All three now get `httptest`-server-backed tests, same
`option.WithBaseURL` pattern as the shared suite above: text
responses, tool-call rounds, `AppendToolResults` folding into the
next request, `ForceFinalResponse` stripping tools and appending the
nudge message, streaming via a text-delta handler, and a plain API
error passing through wrapped but not misclassified as a safety
violation.

Anthropic's SSE stream decoder dispatches on the SSE `event:` line
rather than a type field embedded in the JSON body (unlike OpenAI's
and Gemini's Chat Completions streams), so Claude's streaming test
needed its own two-line `event:`/`data:` frame helper. Claude's
beta-MCP path (used when an MCP server is configured) is a
near-complete parallel implementation of the same logic against Beta
SDK types, so it gets its own equivalent test file.

Per-function coverage in these three files: 0% or 30-50% average
before, now at or near 100% except for a handful of narrow branches
not exercised by these tests (e.g.
`claudeBetaImageBlocksFromUserImages`'s non-image-generating-tool-call
branches).

## Original scope note — claude/openai/gemini adapters
The handoff doc's checkpoint 3 named all eight provider files as
"near-identical." That's true of the five covered by the shared
suite above, but not of `claude_adapter.go`, `openai_adapter.go`, or
`gemini_adapter.go` — those already had substantial dedicated test
files, and their previously-remaining uncovered functions were
bespoke, provider-specific SDK paths, not simple wrapper methods a
shared suite could cover. That gap is what the follow-up section
above closes.

## Coverage note
`go-unit` moved 33.6% → 34.3% → 35.7% across the two commits in this
PR (checkpoint 2 left it at 33.6%, originally 33.4% before checkpoint
1). This checkpoint's ≥48% target assumed all eight adapter files
were equally cheap to cover; even with the claude/openai/gemini
follow-up included, the total movement is well short of that number
because these files carry a bounded number of statements against a
much larger whole-codebase `-coverpkg` denominator. Flagging rather
than force-fitting unrelated work to hit the number.

## Test plan
- [x] `go test ./internal/agent/provider/...` passes
- [x] `make test-ci` passes (full suite green)
- [x] `make coverage-summary PROFILE=coverage/go-unit.out` →
      `35.7%` (from 33.6% at the start of this PR)
- [x] `gofmt`/`go vet` clean

### DIFF
--- a/internal/agent/provider/claude_adapter_beta_call_test.go
+++ b/internal/agent/provider/claude_adapter_beta_call_test.go
@@ -1,0 +1,225 @@
+package provider
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	anthropic "github.com/anthropics/anthropic-sdk-go"
+	"github.com/stretchr/testify/require"
+)
+
+// newTestClaudeMCPAdapter builds a ClaudeAdapter in beta-MCP mode (useBetaMCP),
+// the path Call/ForceFinalResponse/AppendToolResults take when an MCP server is
+// configured. Same wire format as the plain Messages API (id/type/role/model/
+// content/stop_reason/usage), just routed through client.Beta.Messages.
+func newTestClaudeMCPAdapter(baseURL string) *ClaudeAdapter {
+	provider := NewClaudeProviderWithBaseURL("test-key", baseURL, nil, nil)
+	params := anthropic.MessageNewParams{
+		Model:     "test-model",
+		MaxTokens: 100,
+		Messages: []anthropic.MessageParam{
+			anthropic.NewUserMessage(anthropic.NewTextBlock("hi")),
+		},
+	}
+	mcp := &ClaudeMCPConfig{
+		Servers: []anthropic.BetaRequestMCPServerURLDefinitionParam{
+			{Name: "mcp-a", URL: "https://mcp.example.com"},
+		},
+	}
+	return NewClaudeAdapter(provider, params, nil, false, mcp, nil)
+}
+
+func TestClaudeAdapter_BetaMode_Call_TextResponse(t *testing.T) {
+	srv := jsonHandlerServer(t, claudeMessageTextJSON("bmsg_1", "hello beta claude"))
+	defer srv.Close()
+
+	a := newTestClaudeMCPAdapter(srv.URL)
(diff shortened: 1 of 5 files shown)


GoriVix (approved):
LGTM. Will merge as stack

PR #16, opened 2026-08-26
Cover memory.go and chatmessage.go datastore paths

unlikelyzero:
## Summary

Follow-up tranche on the same shared sqlite harness introduced in
oss/unit-tests-datastore (#15), continuing the datastore coverage push
past that PR's shortfall:

- `internal/datastore/memory.go`: 21 previously zero-coverage functions
  now covered (CRUD, batch create, pin/update, chat-summary upsert/get,
  export, related-memory lookup, and the pure-logic helpers). Two
  functions (`GetRelatedMemories`/`GetRelatedSummaryMemories`) rely on a
  PostgreSQL-only pgvector operator sqlite can't parse; their tests
  exercise the query setup and error path, with the shared similarity
  scoring covered directly via `embeddingCosineRelevance`.
- `internal/datastore/chatmessage.go`: all 15 previously zero-coverage
  functions now covered (CRUD, pagination/filters, generation-expression
  and checkpoint/context-breakdown setters). Along the way, replaced six
  pre-existing tests in `chatmessage_test.go` that only re-implemented
  filter-validation logic inline and never called into the package —
  they contributed nothing to this file's coverage despite existing.

`go-unit` coverage: 39.5% -> 42.1%.

Stacked on #15.

## Test plan

- [x] `go test ./internal/datastore/... -race` green
- [x] `gofmt -l` / `go vet` clean
- [x] `make test-ci` + `make coverage-summary` confirm 42.1%

PR #18, opened 2026-08-26
Cover job.go and mcpserver.go, fix personality-media job finder

### LINKED ISSUE #17: FindActivePersonalityMediaJob errors instead of returning newest job
## Summary

`Datastore.FindActivePersonalityMediaJob` (`internal/datastore/job.go`) is
documented as returning "the newest non-terminal personality background job"
for a user. It filters `job_type` across three distinct values
(`expression_grid`, `personality_portrait`, `personality_generation`) and
orders by `created_at` descending, but terminated the query with
`.Only(ctx)` instead of `.First(ctx)`.

ent's `Only()` returns a `*NotSingularError` whenever more than one row
matches the filter, regardless of the `Order` clause — it does not fall back
to picking the first (newest) row. Since the job-type filter spans three
distinct types, it's entirely normal for a user to have two of them active
concurrently (e.g. an `expression_grid` job and a `personality_portrait` job
both mid-flight). In that case the call returned `"ent: job not singular"`
instead of the newest job, contradicting its own doc comment.

## Fix

Use `.First(ctx)` instead of `.Only(ctx)` — it respects the ordering and
returns ent's standard `NotFoundError` (already handled via
`ent.IsNotFound`) when nothing matches.

## Found via

Coverage work on `internal/datastore/job.go` — a new unit test creating two
concurrently-active jobs of different types surfaced the mismatch between
the doc comment and the actual behavior.

unlikelyzero:
## Summary

Follow-up tranche on the shared sqlite harness, stacked on #16:

- `internal/datastore/job.go`: all 24 functions now covered (CRUD,
  status/progress updates, draft-delta handling, cancel/fail finalize
  paths and their tx helpers, and the three `FindActive*`/`FindLatest*`
  finder queries).
- Fixes a real bug found while writing those tests:
  `FindActivePersonalityMediaJob` filters `job_type` across three
  distinct job types and orders by `created_at desc`, but terminated the
  query with `Only(ctx)` instead of `First(ctx)`. ent's `Only()` errors
  whenever more than one row matches, regardless of ordering — so a user
  with two of those job types active concurrently (e.g. an
  `expression_grid` and a `personality_portrait` job both mid-flight) got
  `"ent: job not singular"` instead of the newest job, contradicting the
  function's own doc comment. Fixes #17.
- `internal/datastore/mcpserver.go`: all 13 previously zero-coverage
  functions now covered (CRUD, auth-token encryption round-trip,
  ritual/chat linking, availability listing). No bugs found in this file.

`go-unit` coverage: 42.1% -> 43.8%.

## Test plan

- [x] `go test ./internal/datastore/... -race` green
- [x] `gofmt -l` / `go vet` clean
- [x] `make test-ci` + `make coverage-summary` confirm 43.8%

### DIFF
--- a/internal/datastore/job.go
+++ b/internal/datastore/job.go
@@ -870,7 +870,7 @@ func (d *Datastore) FindActivePersonalityMediaJob(ctx context.Context, userID uu
 		).
 		Order(job.ByCreatedAt(sql.OrderDesc())).
 		WithOwner().
-		Only(ctx)
+		First(ctx)
 	if ent.IsNotFound(err) {
 		return nil, nil
 	}
--- a/internal/datastore/job_test.go
+++ b/internal/datastore/job_test.go
@@ -550,17 +550,13 @@ func TestFindActivePersonalityMediaJob_ReturnsSoleActiveJob(t *testing.T) {
 	require.Equal(t, createdNewer.ID, got.ID)
 }
 
-// TestFindActivePersonalityMediaJob_MultipleActiveJobsErrors documents a real bug: the doc
-// comment on FindActivePersonalityMediaJob (job.go) promises "the newest non-terminal
-// personality background job," and it does apply an Order-by-created_at-desc, but the query
-// terminates with .Only(ctx) rather than .First(ctx)/.Limit(1). ent's Only() errors whenever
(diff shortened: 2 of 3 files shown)


GoriVix (approved):
LGTM

PR #21, opened 2026-08-26
ci: Compact e2e status comments with vs-main baseline and tested SHA

unlikelyzero:
Reworks the sticky E2E PR comments into a tighter, more informative view.

## What changed

**One compact table per suite group**, rendered by a new shared script (`web/app/e2e/scripts/report-summary.mjs`) instead of inline jq. Rows are meaningful suites, not raw Playwright project names:

- **chrome** — `chromium-desktop` functional specs
- **mobile** — `chromium-mobile` + `webkit-mobile`
- **visual** — `tests/visual/` specs across projects

The auth `setup` project is excluded from counts, and a suite that didn't run says why inline (`add the e2e-full label`, `skipped — no changes under web/app/`) instead of separate Scope/Visual paragraphs.

**vs-main baseline column.** Main-branch runs (merge to main, the nightly, manual dispatch) upload a small `e2e-summary-{mock,local-llm}` artifact; PR runs fetch the newest one via the artifacts API and render a `main` column plus a link to the baseline run. The fetch fails open — no baseline just drops the column. `e2e-local-llm.yml` had no main-branch trigger at all, so it gains a nightly (07:30, offset from the mock nightly) to produce that baseline; real inference stays opt-in on PRs.

**Tested SHA.** Each comment names the linked PR head commit it actually tested — some of these runs are label-triggered rather than per-push, so the comment can lag the branch.

**Working report links.** `Full HTML report` now uses upload-artifact's `artifact-url` output (the report upload moved ahead of the comment build), a direct link to the artifact instead of the `#artifacts` anchor.

**Separate comments for mock vs local LLM.** They trigger differently (per-push vs `run-local-model` label), so the local-LLM run now maintains its own sticky comment via a new `comment-key` input on the `e2e-status-comment` action, instead of a section of the mock comment.

**Bug fix in passing:** `merge_comment.py` shelled out to `gh`, which the e2e container image doesn't ship — posting from `e2e-local-llm.yml`'s container job would fail, hidden behind that job's `continue-on-error`. It now uses stdlib urllib. The baseline fetch uses curl+jq+python3 for the same reason.

## Example rendering

> ### E2E · mock LLM ✅
> **100 passed**, 7 skipped · 443s · [`abc1234`](https://example.invalid)
>
> | Suite | This PR | `main` |
> |---|---|---|
> | chrome | ✅ 96/96 | 96/96 |
> | mobile | – add the `e2e-full` label | 190/192 ❌ |
> | visual | ✅ 12/12 | 12/12 |
>
> [HTML report](https://example.invalid) · [run](https://example.invalid) · main baseline: [`def5678`](https://example.invalid) (nightly)

## Notes

- The vs-main column won't appear on PRs until the first main run after merge uploads a baseline artifact.
- Script verified against synthetic reports covering pass/fail/flaky/skipped, baseline/no-baseline, and desktop-only runs.

### DIFF
--- a/.github/actions/e2e-status-comment/action.yml
+++ b/.github/actions/e2e-status-comment/action.yml
@@ -1,13 +1,20 @@
 name: 'Post E2E status comment section'
 description: >-
-  Merges one E2E workflow's results into a single sticky PR comment, as a
-  section keyed by a stable id. Each e2e workflow (mock backend, local-LLM
-  backend, ...) calls this once with its own section id so they land as
-  separate parts of one comment instead of one comment per workflow.
+  Merges one E2E workflow's results into a sticky PR comment, as a section
+  keyed by a stable id. Workflows that share a comment-key land as separate
+  sections of one comment; a workflow with its own comment-key (e.g. the
+  opt-in local-LLM run, which triggers on a different cadence) gets its own
+  sticky comment that updates independently.
 inputs:
   pr:
     description: 'Pull request number to comment on'
     required: true
+  comment-key:
+    description: >-
+      Identity of the sticky comment this section belongs to. Sections
+      sharing a key merge into one comment; a distinct key gets its own.
+    required: false
+    default: 'e2e-status-comment'
   section:
     description: "Stable id for this workflow's section, e.g. mock, local-llm"
     required: true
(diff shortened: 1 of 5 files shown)


GoriVix (approved):
LGTM

PR #23, opened 2026-08-26
Clarify model-independent continuity in README

dahmonson-ux:
## Summary

Clarifies WhatIff's core architectural differentiator near the top of the README: persistent AI state stays in the WhatIff deployment while supported model providers remain interchangeable.

## Why

The existing README accurately described the feature set, but the opening could be read as another multi-provider AI chat frontend. This change makes the continuity and ownership architecture explicit without expanding the scope of the README.

## Changes

- Reframes the opening description around provider-independent continuity and memory
- Explains that conversations, memory, personalities, tools, and scheduled agents live in the WhatIff stack
- Clarifies that model selection is independent from the persistent continuity layer

No implementation behavior changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@
 
 # WhatIff
 
-**An open-source, self-hostable AI assistant with memory, personalities, agents, and tools.**
+**An open-source, self-hostable AI assistant where your continuity and memory are not tied to a single model provider.**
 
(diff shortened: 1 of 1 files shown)


GoriVix (approved):
Good change! README definitely could use some work. we threw it together pretty fast!

PR #26, opened 2026-08-27
feat(chat): message bookmarks + long-thread navigator

GoriVix:
Port of chat-app PR  to what-iff.

Adding "bookmarks": Bookmarks allow users to 'tag' specific messages in a conversation and then use the new, top menu to auto-scroll back to that point in the conversation. Allowing users to find things in long conversations where scrolling or search are not practical.

Also done in this PR: updated our logic around chat rendering to prevent the chat window from resetting to bottom when user leaves + returns and limit the 'jarring' shifts when loading additional pages of history.

- New `bookmarked` bool on ChatMessage (ent schema + auto-migrate index)
- PATCH /chat/{chatId}/chat-message/{messageId}/bookmark and GET /chat/{chatId}/bookmarks
- BookmarkSnippet / BookmarkPageDefault helpers in models
- ListChatMessageBookmarks (complete, for UI navigator) + ListChatMessageBookmarksPage (bounded, for agent tools)
- find_context `bookmarks` mode + bookmark:<chat>:<msg> fetch-target prefix so agents can scan compact labels before fetching a body
- refactored `recallConversationTarget` helper shared by conversation and bookmarks modes
- Frontend: star toggle on every message bubble (optimistic, reverts on error)
- ThreadBookmarksComponent: gold pill button in thread header showing count, dropdown navigator
- Jump logic: loadOlderMessagesUntil pages back until the target is loaded, scrollToMessage retries across rAF frames and flashes the bubble
- reconcileLatestPage for non-destructive tab-return sync (no scroll position loss); syncActiveThread replaces reloadActiveThread
- Anchor-based scroll restoration after older-page prepend (replaces rawscrollHeight delta); infinite auto-load as user scrolls toward top
- openapi.yaml updated with both endpoints and ChatMessageBookmark schema

### DIFF
--- a/ent/schema/chatmessage.go
+++ b/ent/schema/chatmessage.go
@@ -65,6 +65,9 @@ func (ChatMessage) Fields() []ent.Field {
 		field.JSON("context_breakdown", &modeltypes.ContextBreakdown{}).
 			Optional().
 			Comment("jsonb snapshot (modeltypes.ContextBreakdown) of the model-context segment/token composition captured at generation time for this assistant message; used by the Context X-ray UI. jsonb (not text) keeps DB-side JSON queries open; a version field inside guards shape evolution."),
+		field.Bool("bookmarked").
+			Default(false).
+			Comment("User-flagged bookmark for navigating long threads; either origin can be bookmarked."),
 	}
 }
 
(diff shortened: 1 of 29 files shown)


GoriVix:
Broken test is unrelated race-condition issue. Fixing in a separate PR...

dahmonson-ux (approved):
Reviewed. LGTM. The remaining mock-backend E2E failure is documented as an unrelated race condition being handled separately.

PR #27, opened 2026-08-27
Fix summarizer crash on OpenAI-unsupported images; guard pod against panics

GoriVix:
## Summary

The checkpoint summarizer crashed on images OpenAI won't accept. Some vendors (Anthropic, Gemini) allow image types/filenames OpenAI's Responses API rejects, so when a Claude-model chat reached its checkpoint and rendered those images into the OpenAI summary request, OpenAI returned `400 Bad Request` (`Expected image type ... but got .JPG`) and the whole summary failed. The `.JPG` is OpenAI validating a stored file's **filename extension case-sensitively** on a `file_id` reference — the media type (`image/jpeg`) is valid, the uploaded `.JPG` name is not.

- **Strip OpenAI-unusable images before the summary** (`SanitizeImagesForOpenAIInput`): drop unsupported media types, and render survivors from raw bytes (clearing `FileID`) so we never rely on a stored filename OpenAI might reject. Only affects the Claude checkpoint path; OpenAI chats summarize via `PreviousResponseID` and always carry raw bytes.
- **Normalize uploaded filename extensions to lowercase** so future `file_id` references validate at the source; the strip above is the safety net for files already stored with bad-case names.
- **Recover panics at the async chat-message goroutine root** so a failure fails just that job instead of crashing the process, mirroring the existing `agentjob_run.go` guard.

## Test plan

- [x] `make pre-commit` (fmt, vet, tidy, test, build) — gofmt/vet clean on touched files; `go build ./...` and `go test ./internal/agent/...` pass
- [ ] Frontend lint/test/build (only if `web/` changed) — N/A, backend only
- [x] Manually verified: added regression tests reproduce the exact crash shape (a turn whose image has both a `file_id` and an unsupported sibling) and assert the built summary params surface no `file_id` image and only a normalized data URL; plus sanitizer edge cases, filename normalizer, and the panic guard (contained even when the failed-status DB write errors)

## Checklist

- [x] Public defaults still work without hosted services (local JWT auth, unlimited usage) — no auth/usage surfaces touched
- [ ] Updated `docs/ARCHITECTURE_SUMMARY.md` / the affected `_PACKAGE_SUMMARY.md` if this touches system boundaries — no boundary changes
- [ ] Regenerated the e2e SDK if `openapi.yaml` changed — N/A

### DIFF
--- a/internal/agent/conversation_summary.go
+++ b/internal/agent/conversation_summary.go
@@ -61,7 +61,10 @@ func buildCheckpointSummaryParams(userID uuid.UUID, existingSummary string, src 
 			OfString: openai.String(updatePrompt),
 		}
 	case src.ModelContext != nil:
-		items := provider.RenderOpenAIInputItems(src.ModelContext)
+		// Strip images OpenAI won't accept (unsupported media types, and file_ids whose
+		// stored filename extension OpenAI rejects case-sensitively, e.g. ".JPG"). Without
+		// this, a single bad image fails the whole checkpoint summary with a 400.
+		items := provider.RenderOpenAIInputItems(provider.SanitizeImagesForOpenAIInput(src.ModelContext))
 		if trimmed := strings.TrimSpace(src.AssistantReply); trimmed != "" {
 			items = append(items, responses.ResponseInputItemParamOfMessage(trimmed, provider.RoleAssistant))
 		}
--- a/internal/agent/conversation_summary_test.go
+++ b/internal/agent/conversation_summary_test.go
@@ -85,6 +85,48 @@ func TestBuildCheckpointSummaryParams_ExplicitMode(t *testing.T) {
 	require.Len(t, items, 4) // history user, current user, assistant reply, update prompt
 }
 
+// TestBuildCheckpointSummaryParams_StripsOpenAIUnsafeImages reproduces the checkpoint
+// summary crash: a Claude-chat turn carried an image whose OpenAI file_id has an uppercase
+// ".JPG" extension (rejected case-sensitively) plus a media type OpenAI does not support,
+// which returned a 400 from the summarizer. The built params must never surface a file_id
+// image or an unsupported media type.
+func TestBuildCheckpointSummaryParams_StripsOpenAIUnsafeImages(t *testing.T) {
+	t.Parallel()
+
+	mc := &provider.ModelContext{}
+	// Supported type carrying both a file_id (the ".JPG" landmine) and raw bytes: kept, but
+	// must render as a data URL, never as the file_id reference.
+	mc.AppendHistoryTurn(provider.RoleUser, "look at this", []provider.UserMessageImage{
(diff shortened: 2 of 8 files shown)


GoriVix:
Same broken test as #26 fix in separate PR

dahmonson-ux:
One edge case I think is worth checking before merge: `openAISupportedImageMediaTypes` explicitly accepts `image/jpg`, but `normalizeImageMediaType()` appears to pass that value through unchanged. `image/jpg` is non-standard; the standard MIME type is `image/jpeg`.

That matters because the sanitizer treats `image/jpg` as OpenAI-safe and then forces the image onto the data-URL path. If OpenAI rejects `data:image/jpg;base64,...`, we can still get the same class of 400 this PR is intended to prevent, just from a MIME-type alias instead of the filename extension.

Possible fix: canonicalize `image/jpg` to `image/jpeg` before building the data URL (ideally in `normalizeImageMediaType()`), and add a regression test that feeds an `image/jpg` image through `SanitizeImagesForOpenAIInput` / rendering and asserts the resulting URL starts with `data:image/jpeg;base64,`. Alternatively, if OpenAI is confirmed to accept `image/jpg`, a test documenting that assumption would close the question.

GoriVix:
> One edge case I think is worth checking before merge: `openAISupportedImageMediaTypes` explicitly accepts `image/jpg`, but `normalizeImageMediaType()` appears to pass that value through unchanged. `image/jpg` is non-standard; the standard MIME type is `image/jpeg`.
> 
> That matters because the sanitizer treats `image/jpg` as OpenAI-safe and then forces the image onto the data-URL path. If OpenAI rejects `data:image/jpg;base64,...`, we can still get the same class of 400 this PR is intended to prevent, just from a MIME-type alias instead of the filename extension.
> 
> Possible fix: canonicalize `image/jpg` to `image/jpeg` before building the data URL (ideally in `normalizeImageMediaType()`), and add a regression test that feeds an `image/jpg` image through `SanitizeImagesForOpenAIInput` / rendering and asserts the resulting URL starts with `data:image/jpeg;base64,`. Alternatively, if OpenAI is confirmed to accept `image/jpg`, a test documenting that assumption would close the question.

Fair. OpenAI does announce they support .jpg, from our error logs:
```
400 Bad Request {\n \"message\": \"Invalid input: Expected image type to be a supported format: .jpeg, .jpg, .png, .gif, .webp but got .JPG.\"
```

Unfortunately our test suites do not make live calls to APIs (for cost reasons, etc) so we dont have a great way to validate if they are, without just encoding into the test that they would

dahmonson-ux (approved):
Reviewed the current PR. The image sanitization, lowercase upload-extension normalization, and async panic containment all look sound. The `image/jpg` concern is acceptable given the documented API contract, and the remaining mock E2E failure is the same unrelated race-condition issue already identified for a separate fix. LGTM.

PR #29, opened 2026-08-27
Fix thread manager archive search failures

GoriVix:
## Summary
- ignore stale thread-list refresh responses and wait for filtered-list responses in the E2E POM
- make Postgres tag searches tolerate legacy non-array JSON, avoiding the `/api/chat` 500 that prevented archived threads from appearing
- cover response ordering in `ThreadListService`

## Test plan
- [x] `ENV=test LLM_BACKEND=mock OPENAI_API_KEY=dummy-ci-key go test ./internal/datastore`
- [x] `npm test -- --watch=false --include='src/app/core/services/thread-list.service.spec.ts'`
- [x] `npm run lint:e2e`
- [x] `make ci-local ARGS=\"--only e2e\"`

Made with [Cursor](https://cursor.com)


GoriVix:
MErging to fix broken test

PR #30, opened 2026-08-27
test(e2e): expand responsive layout contracts

dahmonson-ux:
## Summary

Adds tests only. No production UI/CSS changes.

This PR expands the responsive E2E contract so future layout fixes are judged by user-visible behavior rather than a specific CSS implementation.

### Coverage added

- Exercises narrow widths: 320, 360, 375, 390, 412, 768, and 1023 CSS px, plus the 1024px desktop breakpoint where relevant
- Adds reusable geometry assertions for viewport containment, parent/child containment, balanced horizontal spacing, and horizontal overflow
- Verifies the mobile navigation control does not overlap the Memories heading/subtitle or the Jobs heading/Create action
- Verifies Memories header copy/actions, filters, sort controls, and cards stay usable and in-bounds
- Verifies the Mode create dialog and its visible form controls remain contained across mobile widths
- Adds a constrained-height Mode dialog case to ensure long modal content scrolls inside the dialog instead of escaping the viewport
- Verifies the Integrations connector search input and Search button remain contained and horizontally balanced within their search row
- Applies content pressure to the Integrations search input and verifies it does not introduce page-level horizontal scrolling
- Uses Playwright trial actions where appropriate to verify enabled controls remain actionable

The assertions deliberately do **not** prescribe whether a future fix uses padding, flex/grid changes, wrapping, width constraints, or another layout approach.

## Current expected result

This is intentionally a tests-only PR. Responsive assertions may expose existing production layout defects on the current implementation; production fixes belong in the stacked fix PR.

## Main files

- `web/app/e2e/helpers/responsive-layout.ts`
- `web/app/e2e/poms/memories.page.ts`
- `web/app/e2e/tests/functional/memory/responsive-layout.spec.ts`
- `web/app/e2e/tests/functional/mode/mode-edit.spec.ts`
- `web/app/e2e/tests/functional/integrations/integrations.spec.ts`

### DIFF
--- a/web/app/e2e/helpers/responsive-layout.ts
+++ b/web/app/e2e/helpers/responsive-layout.ts
@@ -1,0 +1,78 @@
+import type { Locator, Page } from '@playwright/test';
+import { expect } from '@playwright/test';
+
+export const RESPONSIVE_WIDTHS = [320, 360, 375, 390, 412, 768, 1023] as const;
+export const DEFAULT_VIEWPORT_HEIGHT = 900;
+export const LAYOUT_TOLERANCE_PX = 1;
+
+export type Rect = NonNullable<Awaited<ReturnType<Locator['boundingBox']>>>;
+
+export async function rect(locator: Locator, label: string): Promise<Rect> {
+  const box = await locator.boundingBox();
+  expect(box, `${label} should have a rendered bounding box`).not.toBeNull();
+  return box as Rect;
+}
+
+export function expectInsideViewport(
+  box: Rect,
+  viewportWidth: number,
+  label: string,
+  viewportHeight?: number,
+): void {
+  expect(box.x, `${label} should not extend past the left viewport edge`).toBeGreaterThanOrEqual(-LAYOUT_TOLERANCE_PX);
+  expect(box.x + box.width, `${label} should not extend past the right viewport edge`).toBeLessThanOrEqual(
+    viewportWidth + LAYOUT_TOLERANCE_PX,
+  );
+
+  if (viewportHeight !== undefined) {
+    expect(box.y, `${label} should not extend past the top viewport edge`).toBeGreaterThanOrEqual(-LAYOUT_TOLERANCE_PX);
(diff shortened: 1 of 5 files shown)


dahmonson-ux:
This PR is intentionally tests-only.

The issue that exposed the gap is a mobile layout regression where the global navigation button can overlap routed page content. On the Memories page, the title can render underneath the hamburger at narrow widths even though the existing E2E test still passes because it only checks that the heading is technically "visible."

That distinction matters: an element can exist in the DOM, have a rendered bounding box, and satisfy "toBeVisible()" while still being partially covered or pushed outside the usable viewport.

The new responsive contract is meant to close that gap without prescribing a specific UI fix.

It exercises the Memories page at 320, 360, 375, 390, 412, 768, 1023, and 1024 CSS px and checks user-visible invariants instead of CSS implementation details:

- navigation controls cannot overlap the page heading/subtitle
- header copy and actions cannot overlap each other
- key controls and memory cards must remain inside the viewport
- the page cannot require unintended horizontal scrolling
- enabled controls must remain actionable, not merely visible
- the 1023/1024 mobile-to-desktop breakpoint is explicitly covered

The important part for future integration is that these tests do not require any particular fix. The UI can be repaired by adding spacing, wrapping controls, stacking the header, restructuring the mobile shell, moving the navigation control, or using another approach entirely. As long as the resulting interface remains unobstructed and usable, the tests pass.

That gives future PRs a behavioral guardrail rather than freezing the current layout implementation. It should also make this class of responsive regression much harder to reintroduce while the frontend continues evolving.

GoriVix (commented):
Seems fine. @dahmonson-ux 2 of the new tests are failing, if you could get those fixed up. Otherwise I can tidy it up tomorrow and ship

dahmonson-ux:
@GoriVix won't pr 31 is green 30 will be red since 30 its test harness and 31 is bug fixes.

GoriVix:
@dahmonson-ux  ok usually you want each PR to be green independently but with stacks i suppose it doesnt matter aton

GoriVix (approved):

PR #56, opened 2026-08-29
Fix imported conversation message ordering

### LINKED ISSUE #35: Import archives flipping conversation order
### Area

API / backend (Go)

### What happened

For some bulk imports from other providers, the ordering of messages is swapping from user -> assistant to assistant -> user, which makes revisiting the old threads incredibly annoying

### What you expected

Not that to happen

### Steps to reproduce

Not entirely clear, noticed from reviewing some of my old chats. Possibly has to do with how the importer is putting things into the DB (ordering). Also need to see if we can fix existing imports that were broken this way -- they should be notable because the chat messages will not be in timestamp order.

### Version or commit

main

### Environment

_No response_

### Additional context

_No response_

dahmonson-ux:
## Summary

Fixes #35.

The OpenAI and Anthropic parsers preserve transcript sequence, but imported messages are later retrieved by `(sent_at, id)`. Provider archives can contain equal timestamps, missing timestamps that share the same fallback, regressing timestamps, or source timestamps that differ by less than the datastore can represent. In those cases persisted timestamps can tie and UUID ordering can override the original user/assistant sequence.

The fix enforces the ordering invariant at the shared persistence boundary used by all import providers:

- Imported messages are persisted in source-transcript order.
- PostgreSQL timestamp precision is treated as one microsecond (`importedMessageOrderStep = time.Microsecond`).
- The first imported timestamp is preserved in UTC.
- For each later message, `normalized_sent_at = max(source_sent_at, previous_normalized_sent_at + importedMessageOrderStep)`.
- Source timestamps that are already at least one datastore-supported precision step later are preserved exactly.
- Equal, missing/fallback, regressing, or sub-microsecond-separated values are advanced only as far as needed to maintain a datastore-distinguishable monotonic sequence.
- The comparison uses the immediately preceding **normalized** timestamp, not the preceding raw source timestamp.

This normalization is scoped to imported-message persistence; ordinary authored messages are unchanged.

## Validation

Boundary tests now cover the persistence-order contract directly:

- Equal and regressing timestamps are normalized into strict transcript order.
- Sub-microsecond source differences (for example `T`, `T+500ns`, `T+750ns`) normalize to `T`, `T+1µs`, `T+2µs`, preventing persistence-time collapse.
- Precision-safe increasing source timestamps remain unchanged.
- Reapplying normalization to an already-normalized sequence is idempotent.
- Existing OpenAI/Anthropic import tests verify source role order is preserved by their parsers.

A passing regression test is evidence for these reproduced conditions, not proof that every possible archive-ordering bug is eliminated.

## Existing broken imports

This PR also includes a deliberately narrow startup repair for recoverable historical exact-timestamp ties. The repair only considers archived imported conversations with a reliable import identity, exactly one User and one Assistant at the tied timestamp, no associated agent job, and no surrounding timestamp collision. It moves only the Assistant by one production-representable precision step and is idempotent on reapplication.

The repair abstains on ambiguous multi-message groups, tool/system/job-generated cases, collisions, concurrent changes, and history that cannot be causally reconstructed from a reliable signal. Non-imported and already-correct history is preserved.

## BSD rerun — prior head, BSD-main (17), coding profile

The following evaluation was produced for the earlier head `6fb4e8e2c7faf74c8df547d7415a37e3297b55b4` and is retained as historical evidence only. It is **not** a current-head BSD result after the datastore-precision/idempotence changes above.

Evaluation: `eval-cce4e892-8161-4107-b488-e26868a76d23`.

The prior BSD result should be rerun against the latest PR head before treating those metrics as current.


GoriVix:
@dahmonson-ux  I think for the repair -- if things are at the exact same timestamp it should ALWAYS be safe to treat it user > assistant, based on the dynamics of LLM. It is kinda fundamentally impossible to send a message response at the same moment as you received one. (this is MAYBE possible with jobs but in that case the ordering is actually arbitrary for distributed system reasons)

TIL and interesting though, it is weird that the exports timestamp the assistant messages with the user-message send time, which is kinda just wrong.

GoriVix (approved):
LGTM

We should take a pass through soon and delete a bunch of these startup jobs that have run and arent really needed anymore

PR #57, opened 2026-08-29
Fix scratchpad updates near output limits

### LINKED ISSUE #38: UpdateScratchpad tool fails if response too long
### Area

API / backend (Go)

### What happened

Agents who call the updateScratchpad tool and try to append context have a tendency to hit issues with JSON parsing, due to truncated outputs

### What you expected

Needs to be more type-safe, when output gets truncated, it breaks the entire agent turn.

### Steps to reproduce

As agent to update scratchpad and append context when the scratchpad is already near max size

### Version or commit

main

### Environment

_No response_

### Additional context

_No response_

dahmonson-ux:
## Summary

Fixes #38.

`update_scratchpad` previously required the model to emit the entire replacement scratchpad in a single JSON string. When a personality scratchpad was already large and the agent only wanted to append context, the function-call payload grew with the whole accumulated scratchpad. If provider output was truncated before the JSON closed, `UpdateScratchpadTool` failed immediately at `json.Unmarshal` and could not recover the intended update.

## Fix

The shared provider-neutral tool contract now exposes an explicit `operation` with `append` and `replace` modes.

- `append` is the normal path for adding context. The model sends only the new delta; the server loads the existing scratchpad and appends the delta before using the existing persistence path.
- `replace` preserves intentional whole-scratchpad rewrites.
- Empty/missing `operation` is treated as `replace` inside the handler for compatibility with already-generated calls using the old contract.
- Unknown operations return a structured tool failure.
- Malformed/truncated JSON behavior remains explicit; this PR does not guess missing arguments or attempt to repair arbitrary truncated JSON.

The tool description also tells models to prefer `append` for normal additions, keeping generated argument size proportional to the new context instead of accumulated scratchpad size.

## Test-first validation

The first commit added a regression requiring an explicit provider-neutral `operation` property. On the pre-fix implementation, CI failed exactly that regression with `update_scratchpad should expose an explicit operation property`.

Production changes were added only after that failure was observed. Additional tests cover append-to-existing, append-to-empty, explicit replacement, compatibility fallback, and rejection of unknown operations. Existing invalid-JSON coverage remains unchanged.

BSD was used as a planning/evidence-structure check. It kept provider-adapter truncation and post-parse datastore rejection as alternatives and required regression evidence before treating the contract diagnosis as established.

## Scope / limitation

This removes the need to echo a near-full scratchpad for ordinary additions, which addresses the reproduced output-budget mechanism. A deliberately huge full `replace` operation can still encounter provider output limits, and this PR does not claim to repair every possible provider truncation mode.

Current-head evaluation: `eval-df59ce9e-da09-4c9f-9438-150e91c4e63c`.

The leading frame remains supported: append/replace reduces ordinary update payload size, and the newer OpenAI `response.incomplete` / `max_output_tokens` recovery path converts malformed function-call arguments into a tool error and continues the agent loop using the prior response id; current PR Validation and both E2E workflows are green. The partial frame remains measurable because fixed scratchpad capacity and deliberately huge full-replace operations are broader design limits that the represented recovery path does not remove.


GoriVix:
@dahmonson-ux  I like the idea of adding an 'append' here as that is the pretty standard agent use-case. BUT!!

1) We still need to fix the actual bug problem here -- when an agent uses the tool and their output gets truncated, it is crashing the conversation turn. We should recover cleanly from that. Honestly we probably should do a pass of this with all output based tools so a failed call doesnt blow up the turn, but the scratchpad is generally the one where it happens.
2) The problem with doing append is that our scratchpads have a fixed maximum length and, in general, they tend to be pretty close to that limit at all times. So in order to append, the agent *usually* needs to remove things. 

These are also generally why I am not a huge fan of this tool, it feels to me like a foot-gun for agents in most cases. But I know @csotherden has use-cases for it especially for short threads or when we dont want to wait for a compaction. 

We should fix the crash part then talk about the tool as a group (maybe it should do the whole compaction pipeline or something, idk.)

GoriVix:
Going to close this one for now to clean up our queue some -- needs more work

Issue #37, opened 2026-08-29
Job 'run now' doesn't leave good breadcrumb


GoriVix:
### Area

Web app (web/app)

### What happened

When triggering 'run now' for a job, the button temporarily changes state (<1 second) then reverts. Due to this, it is very hard to tell that the job actually got run.

### What you expected

Unclear, but *probably* should transition to the thread after you hit run now. Ideally also with a "..." to represent the currently running job. HOWEVER we do not show in UI live running job executions, so that would be a new functionality

### Steps to reproduce

click run now

### Version or commit

main

PR #60, opened 2026-08-29
Keep chat job polling through transient connectivity

### LINKED ISSUE #33: "Error processing response" when connectivity is bad.
### Area

Web app (web/app)

### What happened

In the chat window, often times if you leave the screen and come back while processing, the UI will show an "Error processing message" message, when the request actually succeeded. This will go away if you refresh the screen (or sometimes if you just wait a bit and the poller picks it up)

### What you expected

UI should not show "ERROR" if the job does not fail, instead it should retain the "..." and re-poll for the response.

### Steps to reproduce

1. make a request
2. leave the window, wait a minute
3. open window again

### Version or commit

main

### Environment

_No response_

### Additional context

_No response_

dahmonson-ux:
## Summary

Fixes #33.

This starts with a regression test for the reported failure mode: a transient transport error while polling an otherwise-running chat job must not be turned into a terminal UI failure. The poller should remain active and accept a later successful terminal snapshot.

## Diagnosis

`JobService.pollJob` currently routes any `getJob` error into its outer `catchError`, immediately adds `Failed to process message`, terminates the polling observable, and removes the job from the active polling set in `finalize`. That transport failure is not evidence that the server-side job failed.

Issue #37 was inspected alongside this issue because both mention jobs, but its run-now breadcrumb/navigation behavior is a separate UI flow and is not being combined into this PR.

## Test-first evidence

The first commit adds a deterministic regression: the first poll throws a temporary network error and the next poll returns `complete`. The required behavior is that no false error is inserted, the poll remains active after the transport error, and the later completion is observed.

BSD was used before code changes as a bounded diagnosis/action check. It retained backend terminal-state mismatch and stale rendered-error state as alternatives, while current source inspection supports the poll-transport termination path as the direct mechanism.

A passing regression after the production patch will establish the reproduced transport-error case only; it will not prove every connectivity failure mode is handled.

Reran the transient-polling slice under the packaged `coding` engineering profile (`evidence_floor=0.40`, `ppi_base_sensitivity=4.0`) using the canonical implementation-correctness output space. The bounded mechanistic chain is `CERTIFIED_WITH_LIMITS` at structural reliability `0.833333`; evidence mass is `1.0`. The canonical mapping is configured and comparable, but PPI is `UNAVAILABLE` specifically because `INSUFFICIENT_SUPPORTED_STRUCTURES`: the competing/falsification frame did not independently earn positive support. Response policy is `DIRECT` with `MISSING_INFORMATION`, `PPI_UNAVAILABLE`, and `SCOPE_LIMITED` disclosures.

The repository source registry remains `CALLER_SUPPLIED_UNVERIFIED` / `UNVERIFIED_REGISTRY`. A direct BSD `code.github` verification attempt returned `RETRIEVAL_FAILED` because the execution environment could not resolve external DNS, so this rerun does not claim adapter-verified GitHub provenance. Code and regression remain one dependent evidence group; `1.0` is coverage, not truth probability. Persistent-outage/backoff behavior remains outside this slice; CI/runtime verification remains the acceptance gate.

### DIFF
--- a/web/app/src/app/core/services/job.service.spec.ts
+++ b/web/app/src/app/core/services/job.service.spec.ts
@@ -1,7 +1,7 @@
 import type { MockedObject } from "vitest";
 import { TestBed } from '@angular/core/testing';
 import { provideZonelessChangeDetection } from '@angular/core';
-import { of } from 'rxjs';
+import { of, throwError } from 'rxjs';
 import { provideHttpClient, withXhr } from '@angular/common/http';
 import { provideHttpClientTesting } from '@angular/common/http/testing';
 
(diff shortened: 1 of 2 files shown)


GoriVix (approved):
Seems reasonable to me. LGTM. Yeah any getJob call that fails doesnt really tell us anything interesting. 

There might be some 4xx cases where this is gonna cause issues, but we do not generally use 4xx in this flow so it should be fine enough for now.

PR #61, opened 2026-08-29
Give Run Now a durable chat breadcrumb

### LINKED ISSUE #37: Job 'run now' doesn't leave good breadcrumb

dahmonson-ux:
## Summary

Fixes #37.

This PR starts with a regression test for the missing Run Now breadcrumb. When a scheduled agent job is associated with a chat and the Run Now trigger succeeds, the UI should take the user to that chat so there is a durable place to see the resulting turn instead of only flashing the button state for less than a second.

## Diagnosis

The current `runNow()` flow sets `isRunningNow`, calls the API, sets `successMessage = "Job run triggered."`, refreshes the job, and immediately clears `isRunningNow`. The API currently returns only `{ status: string }`, so there is no execution/job identifier available to build a new live-running execution UI without expanding backend state plumbing.

The detail page already has an `openChat()` path for jobs with `chat_id`, so reusing that existing navigation is the narrowest supported breadcrumb. Jobs without an associated chat should remain on the detail page and keep the success acknowledgement.

BSD was run before code changes as a bounded diagnosis/action check. It kept persistent inline feedback and new execution tracking as alternatives, but the existing chat association and navigation path support the narrower chat breadcrumb without inventing new execution state.

## Test-first evidence

The first commit is test-only. It asserts that a successful Run Now on a job with `chat_id` sets that chat as active and navigates to `/chat`. Current main does not do that, so the regression should fail before the production patch.

A later green test will establish this associated-chat breadcrumb behavior only; it will not imply that live scheduled-job execution state is implemented.

Reran the associated-chat breadcrumb slice under the packaged `coding` engineering profile (`evidence_floor=0.40`, `ppi_base_sensitivity=4.0`) using the canonical implementation-correctness output space. The bounded mechanistic chain is `CERTIFIED_WITH_LIMITS` at structural reliability `0.833333`; evidence mass is `1.0`. The canonical mapping is configured and comparable, but PPI is `UNAVAILABLE` specifically because `INSUFFICIENT_SUPPORTED_STRUCTURES`: the competing/falsification frame did not independently earn positive support. Response policy is `DIRECT` with `MISSING_INFORMATION`, `PPI_UNAVAILABLE`, and `SCOPE_LIMITED` disclosures.

The repository source registry remains `CALLER_SUPPLIED_UNVERIFIED` / `UNVERIFIED_REGISTRY`. A direct BSD `code.github` verification attempt returned `RETRIEVAL_FAILED` because the execution environment could not resolve external DNS, so this rerun does not claim adapter-verified GitHub provenance. Implementation and regression evidence remain one dependent group; `1.0` is coverage, not truth probability. The result supports the represented breadcrumb mechanism only and does not certify live execution-state tracking; repository CI/runtime verification remains the acceptance gate.

### DIFF
--- a/web/app/src/app/features/agent-job/agent-job-detail.component.spec.ts
+++ b/web/app/src/app/features/agent-job/agent-job-detail.component.spec.ts
@@ -404,15 +404,20 @@ describe('AgentJobDetailComponent', () => {
             expect(mockAgentJobService.updateAgentJobStatus).not.toHaveBeenCalled();
         });
 
-        it('runs job now for active/paused statuses and refreshes details', () => {
-            const refreshed = { ...mockJob, run_count: 4 };
-            mockAgentJobService.getAgentJob.mockReturnValueOnce(of(mockJob)).mockReturnValueOnce(of(refreshed));
+        it('runs job now without a chat and refreshes details', () => {
+            const noChatJob = { ...mockJob, chat_id: null };
+            const refreshed = { ...noChatJob, run_count: 4 };
+            component.job.set(noChatJob);
+            mockAgentJobService.getAgentJob.mockReturnValue(of(refreshed));
+            mockRouter.navigate.mockClear();
+            mockChatService.setLastChatId.mockClear();
 
-            component.ngOnInit();
             component.runNow();
 
             expect(mockAgentJobService.runNow).toHaveBeenCalledWith('job-1');
             expect(component.job()?.run_count).toBe(4);
+            expect(mockChatService.setLastChatId).not.toHaveBeenCalled();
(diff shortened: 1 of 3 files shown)


GoriVix (approved):
Seems like an improvement, we can dogfood this one a bit and adjust if it feels off.

PR #62, opened 2026-08-29
Normalize uploaded images before provider handoff

### LINKED ISSUE #39: Image normalization
### Area

API / backend (Go)

### Problem

Vendors support different image formats, which leads to frustrating inconsistencies where images will work perfectly fine with one vendor then fail on another. This is particularly notable with anything that works for a vendor then fails on OpenAI as it will cause issues with summarization

### Proposed solution

We should normalize images on upload to a standard size (preserving aspect ratio) and format. This way we know that all images can be sent to vendors without trouble.

### Alternatives considered

_No response_

### Additional context

_No response_

dahmonson-ux:
## Summary

Fixes #39.

Normalize uploaded images at the shared HTTP upload boundary before they are handed to a model provider. The target contract is PNG with a maximum 2048px longest edge, preserving aspect ratio and never upscaling smaller images.

## Diagnosis

`handlerutils.UploadFileAttachment` is the common boundary that buffers multipart uploads before calling the provider. Today it writes the original bytes to the temp file and passes the original image unchanged to `FileAttachmentUploader`. The OpenAI implementation likewise forwards those bytes/MIME directly to its Files API. That leaves image format and dimensions dependent on what the user uploaded, which is exactly the cross-vendor inconsistency described in #39.

BSD was used before changes to bound the fix. It favored normalization at this shared boundary rather than adding provider-specific transformations, because the issue is explicitly about consistent behavior across vendors. Non-image uploads should remain untouched.

## Normalization contract

- output format: PNG
- longest edge: at most 2048px
- aspect ratio preserved
- smaller images are not enlarged
- metadata passed downstream must describe the normalized bytes

PNG is a common supported image format across the vendors considered and avoids lossy text/screenshot degradation and transparency loss. The 2048px ceiling is an application-level normalization choice, not a claim that every provider imposes that exact limit.

## Test-first evidence

The first commit is test-only. It sends a 4000x2000 PNG through the real shared `UploadFileAttachment` boundary and requires the fake provider to receive 2048x1024 image data. Current main forwards 4000x2000 unchanged, so this regression should fail before the production patch.

A later green test will establish the encoded normalization behavior; it will not prove compatibility with every current or future vendor/model.

Reran the shared-boundary normalization slice under the packaged `coding` engineering profile (`evidence_floor=0.40`, `ppi_base_sensitivity=4.0`) using the canonical implementation-correctness output space. The bounded mechanistic chain is `CERTIFIED_WITH_LIMITS` at structural reliability `0.833333`; evidence mass is `1.0`. The canonical mapping is configured and comparable, but PPI is `UNAVAILABLE` specifically because `INSUFFICIENT_SUPPORTED_STRUCTURES`: the competing/falsification frame did not independently earn positive support. Response policy is `DIRECT` with `MISSING_INFORMATION`, `PPI_UNAVAILABLE`, and `SCOPE_LIMITED` disclosures.

The repository source registry remains `CALLER_SUPPLIED_UNVERIFIED` / `UNVERIFIED_REGISTRY`. A direct BSD `code.github` verification attempt returned `RETRIEVAL_FAILED` because the execution environment could not resolve external DNS, so this rerun does not claim adapter-verified GitHub provenance. Implementation and tests remain one dependent evidence group; `1.0` is coverage, not independent vendor verification or truth probability. The result supports the encoded application normalization contract only; provider compatibility beyond the represented regression surface remains outside the certified slice.

### DIFF
--- a/internal/imageutil/imageutil.go
+++ b/internal/imageutil/imageutil.go
@@ -10,7 +10,7 @@ import (
 	"image"
 	_ "image/gif" // register GIF decoder
 	"image/jpeg"
-	_ "image/png" // register PNG decoder
+	"image/png"
 
 	"github.com/google/uuid"
 	"github.com/theimaginaryfoundation/what-iff/internal/storage"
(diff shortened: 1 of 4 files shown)


GoriVix:
## Compass review (local)

Compared `fix/issue-39-image-normalization` with refreshed `origin/main`.

### Major
- `internal/handlers/handlerutils/fileattachment.go:113-140`: normalization reads the complete upload into memory, then rewrites it. Normalize from the existing temp file into a replacement temp file to avoid this extra buffering at the shared upload boundary.

### Minor
- Distinguish malformed-image errors from internal normalization failures where feasible.
- Avoid re-encoding already-conforming PNGs if that preserves the intended format contract.
- Document the deliberate all-inputs-to-PNG normalization contract.
- Add direct edge-case coverage for the shared dimension-scaling helper.

No critical issues found. A fix for the major finding is in progress.

internal/imageutil/imageutil.go
```diff
@@ -10,19 +10,63 @@ import (
...
-const DefaultThumbnailMaxPx = 256
+const (
+	DefaultThumbnailMaxPx   = 256
+	DefaultUploadImageMaxPx = 2048
+)
+
+// NormalizeForUpload decodes an accepted image from input, scales it down when
+// its longest edge exceeds maxPx, and encodes the result as PNG to output.
+// Images that already fit keep their original dimensions; normalization never
+// upscales. Callers must update the stored filename and content type to PNG.
```
GoriVix: Making a change here to have this method manage the file-name since it is a foot-gun to have callers update the name manually (and will just lead to repeated code)

GoriVix (approved):
LGTM.

PR #63, opened 2026-08-29
Add thread references to the next chat message

### LINKED ISSUE #40: fetch_context() improvement -- provide users an easy way to give context to agent
### Area

API / backend (Go)

### Problem

Fetch_context() allows agents to grab old chat logs. HOWEVER to do so they need to grab the thread-id or thread name (and use list). the thread-id is visible in the UI as part of the chat URL, but this is obtuse.

### Proposed solution

Should be some UI element that allows users to 'add that thread to context' -- probably creating a chip on the next message that adds context to agents (User is referencing something in thread='[thread_id]'

### Alternatives considered

_No response_

### Additional context

_No response_

dahmonson-ux:
Fixes #40.

## Diagnosis

`fetch_context()` can retrieve an old thread once the agent knows its thread id/name, but the UI currently gives the user no direct way to hand that reference to the next turn. The chat composer already has a bounded notion of pending next-turn state (attachments and rituals), while the thread manager already exposes row-level actions.

The narrow implementation is therefore:
- add an **Add to context** action to a thread row,
- queue that thread as visible/removable pending composer context,
- serialize an explicit thread id/name reference into the next user turn so the agent can use its existing context-retrieval tools,
- consume the pending thread reference only after a successful send.

This intentionally does **not** add a new backend message schema or make the UI invoke agent tools directly unless later source evidence shows that is required.

## Test-first evidence

The first commit adds a regression requiring each thread row to expose an `Add thread … to context` action and emit the selected thread without opening it. Current production code has no such output/action, so this test is expected to fail before the implementation commit.

## Earlier epistemic boundary

An earlier planning-stage BSD pass classified this as a bounded change. At that stage PPI/evidence-mass were unavailable because the supplied evidence was caller/model-provided rather than independently certified; that was not treated as causal proof. Passing the regression establishes the encoded UI/send behavior, not prove every future `fetch_context()` workflow succeeds.

Current-head evaluation: `eval-546eba9e-36ef-4572-b5e0-9f6f394f7060`.

The encoded Add-to-context queue/removal/next-send path is supported and the current Frontend + local-LLM + mock E2E workflows are green. The competing frame remains live because the maintainer marked the UX pending and requested product review, and because receiving an explicit thread id/name reference is not the same as proving the downstream agent will reliably choose and succeed with `fetch_context()`. BSD therefore keeps the conclusion scope-limited rather than treating the passing mechanics as full product validation.

### DIFF
--- a/web/app/src/app/features/chat/components/thread-list-panel/thread-row.component.ts
+++ b/web/app/src/app/features/chat/components/thread-list-panel/thread-row.component.ts
@@ -1,5 +1,4 @@
-import { CommonModule } from '@angular/common';
-import { AsyncPipe } from '@angular/common';
+import { AsyncPipe, CommonModule } from '@angular/common';
 import { ChangeDetectionStrategy, Component, computed, inject, input, output, signal } from '@angular/core';
 
 import { Chat } from '../../../../core/models/chat.model';
@@ -10,6 +9,7 @@ import { personalityCoverUrl } from '../../../personality/helpers/cover-image.he
 import { personalityAccent } from '../../../personality/helpers/personality-vm.helpers';
 import { thumbnailCircleToImageStyle } from '../../../../shared/ui/avatar/avatar-thumbnail.helpers';
 import { StarIconComponent, TrashIconComponent } from '../../../../shared/ui/icons/icons';
+import { ContextPanelService } from '../../services/context-panel.service';
 
 @Component({
   selector: 'app-thread-row',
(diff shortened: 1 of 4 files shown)


GoriVix:
Pending on this one -- gonna need to see the UX and @ianlimbs  should review it

ianlimbs:
> Pending on this one -- gonna need to see the UX and @ianlimbs should review it

Got it. Thanks, @GoriVix  - Finalizing an @ reference + an "Add Threads (to chat)" feature. Sending HTML pages for review.

ianlimbs:
@GoriVix - Here's the thread reference prototype, also mobile responsive: https://thread-referencing-cd-4141a33a56.netlify.app/


**Destkop Drag to Composer**
image

**Mobile sample: + Threads (same modal when using /threads)**
image
image


HTML Files:
[Chat Add Thread_B w: mobile long press.html](https://github.com/user-attachments/files/32010866/Chat.Add.Thread_B.w.mobile.long.press.html)
[Chat Add Thread _A.html](https://github.com/user-attachments/files/32010867/Chat.Add.Thread._A.html)

1. desktop only - thread drag and drop into composer. Not activated in mobile for cumbersome UX.
- mobile experiment: Open menu, long press on a thread: Activates multi selector w/ cancel, done and number of threads selected (See Add Thread_B) - optional to implement.
2. desktop + mobile - "+" to add threads. Opens a thread selector w/ checkbox, text search + select starred filter
3. desktop + mobile - /thread. The same action, opens the same thread selector w/ checkbox.

Caveat for item 2:
Thread selector can be further modified to add: 
a. Personality filter
b. LLM Model filter
: Recommendation - keep simple text search + starred + checkbox for now.

PR #64, opened 2026-08-29
Explain summary and scratchpad checkpoint changes

### LINKED ISSUE #41: Ask agents to explain changes to summary + scratchpad
### Area

API / backend (Go)

### Problem

We now show the entire old/new for summary + scratchpad. BUT this is not super useful for users because it is a huge block of text, not very easy to tell what actually changed.

### Proposed solution

Add an additional field to scratchpad and summary generation for 'explanation' and use that to populate a snippet in the UI, then have the full text available as an openable modal to prevent text overload on the page. 

### Alternatives considered

_No response_

### Additional context

_No response_

dahmonson-ux:
Fixes #41.

## Diagnosis

Checkpoint history already stores the authoritative before/after conversation summary and personality scratchpad on a `CompactionEvent`. Those snapshots are content-addressed and can be reused by adjacent checkpoints, so a human-readable explanation belongs to the transition/event, not to the snapshot itself.

The existing UI rendered the full old/new text inline, forcing users to infer the meaningful change from two large blocks.

## Test-first evidence

The first commit added a frontend regression with `summary_explanation` and `scratchpad_explanation`. On the unmodified implementation, Frontend PR Validation went red with exactly one failing test (1111 passed / 1 skipped): the explanation text was absent. Type-check/SDK checks and the local-LLM E2E run passed, so this was the intended product regression rather than a setup failure.

## Implementation

- Add summary and scratchpad explanation metadata to the existing compaction event schema/model.
- Persist scratchpad explanation when the event opens and summary explanation when the new summary is attached.
- Generate each explanation as best-effort audit metadata from the authoritative before/after states. Explanation failure does not fail the checkpoint.
- Surface explanation fields in the web model.
- Replace inline giant before/after blocks with short explanation snippets.
- Keep the complete before/after text and existing revert controls in a focused change-details modal.
- Preserve a fallback message for historical checkpoints that predate explanation metadata.
- Add datastore coverage for explanation persistence and frontend coverage that full text remains hidden until the user opens the detail view.

## Tradeoff

This isolated implementation uses an additional small archival inference for each changed summary/scratchpad transition rather than coupling #64 to other in-flight scratchpad-generation work. That keeps the PR independent, but adds archival-call cost/latency at checkpoint time. The explanation is descriptive audit metadata, not an authority on whether the underlying change was correct.

## Earlier BSD boundary

An earlier planning-stage BSD pass classified the implementation plan as BOUNDED. At that stage PPI/evidence mass were unavailable because the supplied source observations were not independently certified by BSD and provenance remained caller-supplied/unverified. That pass was used to constrain scope and expose the snapshot-vs-transition distinction, not as proof that the implementation was correct.

Passing CI establishes the encoded persistence/display behavior, not that every model-generated explanation is accurate or useful.

Current-head evaluation: `eval-8d85e1b6-4b39-450d-87c2-b19f518decf5`.

Current-head diff, GitHub Actions, mergeability, and maintainer-review observations were carried through the exchange receipt sidecar as host-bound `HOST_BOUND_FALLBACK` evidence. In BSD-main (17), distinct bound receipts count fully toward `E_mass` coverage; the host-mediated retrieval path remains separately recorded as `transport_quality=0.75` and is applied once downstream to canonical support. That establishes host-observed retrieval/binding, not that BSD independently proves the semantic interpretation or that the implementation is true/correct.

The transition-metadata mechanism is represented by the current diff and current PR Validation, Frontend PR Validation, local-LLM E2E, and mock E2E are green. The architecture remains genuinely contested because the maintainer said the original intent was to generate explanations inline with the updates themselves, while this head adds a separate archival inference. The PR is also currently non-mergeable, so the equal BSD comparison weight should not be read as an acceptance verdict.

### DIFF
--- a/ent/schema/compaction_event.go
+++ b/ent/schema/compaction_event.go
@@ -53,6 +53,12 @@ func (CompactionEvent) Fields() []ent.Field {
 		field.String("reason").
 			Optional().
 			Comment("Checkpoint decision reason (token budget / message count / etc.)"),
+		field.Text("summary_explanation").
+			Optional().
+			Comment("Short model-generated explanation of what changed in the conversation summary"),
+		field.Text("scratchpad_explanation").
+			Optional().
+			Comment("Short model-generated explanation of what changed in the personality scratchpad"),
 		// FK columns to CheckpointSnapshot, surfaced through the edges below.
 		field.UUID("old_summary_id", uuid.UUID{}).Optional().Nillable(),
 		field.UUID("new_summary_id", uuid.UUID{}).Optional().Nillable(),
--- a/internal/agent/compaction_event_log.go
+++ b/internal/agent/compaction_event_log.go
@@ -35,15 +35,27 @@ func (a *Agent) beginCompactionEvent(
 		return nil
 	}
 
+	scratchpadExplanation := ""
+	if hasScratchpad {
+		if explanation, err := a.explainCheckpointChange(ctx, userID, "personality scratchpad", chatCtx.chat.Scratchpad, newScratchpad); err != nil {
+			a.logger.Warn("failed to explain scratchpad checkpoint change; continuing without explanation",
+				zap.String("chat_id", chatID.String()),
+				zap.Error(err))
+		} else {
+			scratchpadExplanation = explanation
+		}
+	}
+
 	input := models.CompactionEventInput{
-		ChatID:         chatID,
-		Provider:       providerName,
-		Reason:         reason,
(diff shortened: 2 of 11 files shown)


GoriVix:
This one needs a bit of design review. Our original idea was to do this inline with the updates themselves (as asking an agent to explain their process also usually leads to better output results) -- I'll poke at this later

GoriVix:
Going to close this one until we have a design, keep the queue clean. We can revisit!

PR #66, opened 2026-08-29
Let users read uploaded text file contents

### LINKED ISSUE #43: Cannot fetch contents of uploaded text files
### Area

API / backend (Go)

### Problem

When updating text files in an agent RAG, it is impossible to get the existing contents either for backup purposes or to determine what is uploaded.

### Proposed solution

These are accessible now, so we should provide a UI element that lets you fetch and read the text of a file.

### Alternatives considered

_No response_

### Additional context

_No response_

dahmonson-ux:
Fixes #43.

## Diagnosis

Uploaded RAG documents already had owner-scoped attachment lookup and storage text resolution, but the HTTP layer exposed only list + delete and the personality attachment UI had no way to read stored content.

## Implementation

- Add authenticated `GET /file-attachment/{id}/content`.
- Reuse `GetFileAttachment` for owner authorization.
- Reuse `ResolveAttachmentTextContent` rather than exposing object-store details.
- Return readable content as JSON text and explicitly reject attachments without readable stored text.
- Add `getFileAttachmentContent()` to the web client.
- Add a **Read** action to the existing personality RAG attachment list.
- Show returned content in a plain-text `<pre>` viewer with loading, close, and error states.
- Keep uploaded content untrusted: Angular interpolation renders literal markup as text rather than HTML.

## Regression coverage

- Router coverage requires the authenticated content route.
- The personality-page regression loads a real attachment row, clicks `Read`, verifies the correct attachment ID is requested, and verifies returned `<b>literal user text</b>` appears literally with no `<b>` element rendered.

## BSD boundary

BSD was used to keep authorization/storage ownership at the existing backend seam and avoid direct object-store access from the browser; it is not treated as independent proof.

## Validation

Exact current head `23caf9a` passes PR Validation, Frontend PR Validation, local-LLM E2E, and mock-backend E2E.

Current-head evaluation: `eval-dd0d7a86-3b74-41be-a1b0-6bd52ede5e3b`.

The owner-scoped backend read seam and literal-text rendering path remain supported, and current PR Validation, Frontend PR Validation, local-LLM E2E, and mock E2E are green. The long-file UX frame remains live rather than disappearing: the maintainer explicitly noted that a popup from the personality screen may not work well for significantly long files. BSD therefore supports the authorization/rendering mechanism while keeping the product-level conclusion scope-limited.


GoriVix:
Seems reasonable, we are going to have to do a UI pass on this to get it right -- I doubt a popup from within the personality screen is gonna work too well for significantly long files

GoriVix:
I am going to close this one for now -- it was broken when I downloaded it, we can return to the bug later. Trying to clean up our PR list

PR #68, opened 2026-08-29
Add full-screen personality editing flow

### LINKED ISSUE #45: Edit personality should be able to expand to full screen
### Area

Web app (web/app)

### Problem

It is a real pain in the ass to edit personality or scratchpad from the pop-out, due to the text density. That activity really demands full screen interface

### Proposed solution

We actually *have* a full screen personality edit page that appears as a fall back on some paths. Ressurect this, bring it into sync with the modal, and make a flow to expand / contract while editing personality

### Alternatives considered

_No response_

### Additional context

_No response_

dahmonson-ux:
Fixes #45.

## Diagnosis

The app already has a route-backed personality detail page at `/personality/:id`, but it is a different editing surface from the dense edit modal. The route page edits only a subset of the modal's fields, so simply linking to it would preserve the drift issue described in #45 rather than solve it.

## Test-first

The first commit adds a focused regression to the existing personality edit modal spec requiring an explicit full-screen/expand editing affordance. Current production does not render one, so this commit is expected to go red before production changes.

## Intended implementation boundary

- Add a clear expand action from the modal.
- Reuse/shared-state the existing edit surface where practical rather than create another independently drifting form.
- Preserve in-progress edits when expanding/contracting; navigation must not silently discard a dirty draft.
- Bring the full-screen path into parity for the dense personality fields involved in the edit flow.
- Keep this web-app-only; no backend changes unless source tracing proves one is actually required.

## Earlier BSD boundary

An earlier planning-stage BSD pass classified the represented task as BOUNDED while canonical PPI was unavailable at that stage. It was used to lock the engineering scope and alternatives, not to prove which UI architecture is causally correct.

Passing tests establish the encoded expand/contract behavior and state-preservation cases, not prove every possible navigation path is free of state loss.

Current-head evaluation: `eval-8d39de1f-54a7-4188-808e-c7f21acf1992`.

The draft-preserving expand/contract mechanics are represented by the current diff and the Frontend + local-LLM + mock E2E workflows are green. The product boundary remains equally live, though: the maintainer clarified that expansion should stop at the personality panel rather than fill the viewport, while the current head still uses `100vw` / `100dvh` fullscreen styling. BSD therefore keeps the mechanically-correct and partial-correctness frames at equal comparison weight.


GoriVix:
I'm going to do some UI polish on this one, it should only expand out to the personality panel not the entire screen (bad description on my part)

GoriVix:
Tidying this up!

Also fixing the issue where thumbnail expand stops when it hits any wall of the picture (so you can now easily max-size the circle without precise placement)

image

image

Issue #47, opened 2026-08-29
Tool use per model is too coarse


GoriVix:
### Area

API / backend (Go)

### Problem

With our expansion to more model providers, it is now very opaque in the app which providers/models support which tools.

### Proposed solution

We need to expand our UI and backend models to provide more fine-grained information about what tools are available where. (also if they support image vision). UI treatment needs some design work. 

Backend should not expose tools that dont work on that model (wasting tokens). 

This should cover MCP too -- which currently only works on claude and anthropic (although the real fix there is to run our own MCP client)

PR #70, opened 2026-08-29
Add fine-grained model capability metadata

dahmonson-ux:
Refs #47.

## Diagnosis

The current model persistence/API contract exposes only a coarse `tool_support` boolean. That cannot represent vision support, MCP support, or which concrete tool families should be exposed to a model. Issue #47 therefore crosses three separate concerns: capability metadata, runtime tool filtering, and UI presentation.

## Test-first

The first commit adds a regression requiring the model API contract to expose structured capabilities with `tool_calling`, `vision`, `mcp`, and an explicit `tools` list. Production code is intentionally unchanged so CI can establish the missing-contract red state first.

## BSD boundary

The initial BSD pass classified this scope as UNDERDEFINED rather than BOUNDED. That matches the source trace so far: the persisted/API schema is clearly too coarse, but the runtime tool-assembly and provider-specific MCP paths still need to be traced before choosing the enforcement implementation. BSD is being used as a scope/evidence-structure check, not as proof of a particular architecture.

Passing the initial regression establishes the capability-contract slice only; unsupported-tool runtime filtering and UI capability presentation remain separate work under #47.

Current-head evaluation: `eval-e8cdef35-e52f-41d5-9388-8943c9e537e0`.

The current head supports the bounded capability-contract slice across structured API metadata, runtime tool filtering, and model-picker presentation, and current PR Validation, Frontend PR Validation, local-LLM E2E, and mock E2E are green. The partial frame is net-refuted for that represented slice, but provider/model capability assignments are still derived application metadata; per-model provider-side verification remains outside the certified claim.


internal/agent/tools.go
```diff
@@ -80,8 +81,51 @@ func (a *Agent) buildTurnToolPolicy(ctx context.Context, chatCtx *chatContext, u
...
+		policy.disabledTools[agenttools.ToolNameWebSearch] = true
+		return
+	}
+
+	// Native web search is only wired on the OpenAI Responses and native
+	// Anthropic paths. Other providers should not advertise a tool they cannot run.
+	provider := models.ProviderForModel(model.Provider, model.Name)
+	if provider != models.ModelProviderOpenAI && provider != models.ModelProviderAnthropic {
+		policy.disabledTools[agenttools.ToolNameWebSearch] = true
+	}
```
GoriVix: this not being wired to other providers is probably a bug, at least some of them do support web-search (and I think some have even called it before but I could be wrong)

GoriVix: If bug, this is fine but we should file a new issue to add web-search to other providers that support it!

We've also thought about abstracting that tool which we should mention in the ticket -- it would not be hard to do what we do for generate_image(), then we have a consistent bar across providers.

GoriVix:
Working on closing the gaps here on MCP + search. Vision is the main one. I'm going to subsume this with a new PR that just covers vision!

Issue #48, opened 2026-08-29
Improve the tool sidebar


GoriVix:
### Area

API / backend (Go)

### Problem

The tool sidebar currently shows the agent prompt, in unformatted text. it is pretty much unreadable and not human friendly.

### Proposed solution

Add human-useful descriptions to tools and show those in the UI, making it a lot cleaner.

Consider adding human-useful tool names too, although this is more fraught because if the user asks agent to use the 'recall' tool and it is actually find_context(), that will likely confuse agents.

PR #71, opened 2026-08-30
Add human-readable tool descriptions

dahmonson-ux:
Refs #48.

## Diagnosis

The tool catalog carries agent-facing `FunctionToolSpec.Description` values that are intentionally written for model/tool execution. Reusing those prompts as UI copy couples agent instructions to human presentation. PR #71 separates those concerns with `FunctionToolDefinition.HumanDescription` while leaving every agent-facing `Spec.Description` intact.

The review follow-up traced one representative tool end-to-end before changing the UI path:

`FunctionToolDefinition.HumanDescription` → `agent.GetAvailableTools` → `/tools` JSON `description` → Angular `ToolService` → right-sidebar `ContextToolsTab` → `{{ tool.description }}`.

That trace showed there was no serialization-loss bug and no missing production Angular mapping: the sidebar already renders the API `description` field, and #71 already resolves catalog metadata into that field. The missing contract was fallback behavior for runtime/external tools, plus semantic accuracy of several human descriptions.

## Human-description contract

`GetAvailableTools` now owns the presentation precedence rule centrally:

1. trim `HumanDescription`;
2. if it is non-empty, expose it through `/tools.description`;
3. otherwise fall back to the existing agent-facing `Spec.Description`.

This keeps the frontend renderer simple, prevents blank descriptions for runtime/external tools that do not provide `HumanDescription`, and does not make the new field mandatory for plugins or external catalogs. Provider/tool execution continues to use `FunctionToolSpec.Description` unchanged.

## Catalog audit / review corrections

The human copy was checked against the current implementation rather than inferred from tool names:

- `update_scratchpad`: corrected to state that the personality scratchpad persists across conversations using that personality.
- `list`: replaced “and other resources” with the implemented set: models, personalities, skills, files, conversations, jobs, and MCP servers.
- `create_agent_job`: now describes one-time/recurring natural-language scheduling, current-vs-new thread behavior, and optional model/skills overrides. The tool does **not** expose a direct personality-override argument, so the human copy deliberately does not claim one.
- `find_context`: now explains search/retrieval across memories, files, summaries, and conversation history, plus the common `list` → `find_context` workflow without making `list` mandatory.
- `create_memory`: corrected so it does not imply all memory is cross-conversation; the tool can create conversation-scoped or future/global memory.
- `run_subagent`: expanded to reflect the implemented optional personality, model, and skills controls.

This audit supports the specific descriptions inspected here; it is not a claim that every future tool description or capability is permanently complete.

## Regression coverage

Added/updated tests at the actual seams:

- built-in user-toggleable catalog entries must have non-empty human descriptions distinct from agent prompts;
- `GetAvailableTools` prefers trimmed human copy;
- a runtime/external tool with whitespace `HumanDescription` falls back to `Spec.Description`;
- `/tools` handler serialization returns the resolved display metadata;
- the right-sidebar component renders descriptions returned by the tools API.

The frontend production component did not need a behavioral code change because it already renders `tool.description`; its component test now explicitly proves the resolved human text appears in the sidebar.

## Current-main validation

PR head: `b480999d63a16fa29e9caefac0d7ead62eea5fec`.

The branch history itself is still behind `main` because the available connector does not expose a safe PR-branch sync operation. GitHub generated merge ref `602537c8b3769e25f7064c3da39ab99b7d5a1cec`, merging this head into current `main` (`549fb716d508b08fcae27fa2397b92a6d06b34ed`), and the active PR workflows are testing that combined state.

Validated on that current-main merge state:

- PR Validation: **passed** (format, vet, hygiene, tidy, dependency/default checks, race-enabled hermetic Go tests, coverage step, build, and hermetic mock E2E).
- Frontend PR Validation: **passed**. The exact `context-tools-tab.component.spec.ts` passed 3/3; the frontend suite reported 1,125 passing tests with one skipped, and the production Angular build passed.
- The separate long-running E2E mock-backend workflow was still in progress at the time of this update; the hermetic mock E2E inside PR Validation already passed.

## BSD final evaluation — BSD-main (14), coding profile

Final narrow-scope evaluation: `eval-f60045af-1fcc-4865-8b89-85e32925c9c9`.

Evaluated claim: user-facing tool metadata trims and prefers `HumanDescription`, falls back to `Spec.Description` for absent/whitespace runtime descriptions, propagates the resolved value through `/tools` to the existing right-sidebar renderer, and leaves agent-facing `Spec.Description` unchanged. The catalog wording audit was represented separately and limited to the inspected implementations.

- mechanistic chain: `CERTIFIED_WITH_LIMITS`
- structural reasoning reliability: `0.704` (`MEASURED`; not truth probability)
- competing “UI/fallback still missing” chain: `REJECTED`

BSD limitations remain explicit: implementation and regression tests are a dependent evidence group; CI independently executes the current-main merge state but does not independently prove semantic truth of every description; no interactive browser screenshot was captured; and GitHub evidence supplied to this local BSD run is caller-supplied/unverified rather than adapter-verified provenance.


internal/agent/tools/catalog.go
```diff
@@ -5,22 +5,23 @@ import "github.com/openai/openai-go/v3/responses"
...
+	AgentDefault     bool
+	MoodOnly         bool
+	UserToggleable   bool
 }
 
 var functionToolCatalog = []FunctionToolDefinition{
-	{Spec: UpdateScratchpadToolSpec, AgentDefault: true, UserToggleable: true},
-	{Spec: CreateMemoryToolSpec, AgentDefault: true, UserToggleable: true},
-	{Spec: ListToolSpec, AgentDefault: true, UserToggleable: true},
+	{Spec: UpdateScratchpadToolSpec, HumanDescription: "Update the assistant's temporary working notes for the current conversation.", AgentDefault: true, UserToggleable: true},
```
GoriVix: scratchpad is cross conversation so this description is incorrect

internal/agent/tools/catalog.go
```diff
@@ -5,22 +5,23 @@ import "github.com/openai/openai-go/v3/responses"
...
+	{Spec: CreateMemoryToolSpec, HumanDescription: "Save useful information to long-term memory for future conversations.", AgentDefault: true, UserToggleable: true},
+	{Spec: ListToolSpec, HumanDescription: "Browse available conversations, files, jobs, personalities, skills, and other resources.", AgentDefault: true, UserToggleable: true},
```
GoriVix: I think we should make sure this is a full list, the 'and other resources' is not very helpful for users

internal/agent/tools/catalog.go
```diff
@@ -5,22 +5,23 @@ import "github.com/openai/openai-go/v3/responses"
...
 	{Spec: ListMoodsToolSpec, AgentDefault: true, MoodOnly: true, UserToggleable: false},
 	{Spec: ChangeMoodToolSpec, AgentDefault: true, MoodOnly: true, UserToggleable: false},
-	{Spec: RunSubagentToolSpec, AgentDefault: true, UserToggleable: true},
-	{Spec: GenerateImageToolSpec, AgentDefault: true, UserToggleable: true},
-	{Spec: CreateAgentJobToolSpec, AgentDefault: true, UserToggleable: true},
-	{Spec: RecallToolSpec, AgentDefault: true, UserToggleable: true},
+	{Spec: RunSubagentToolSpec, HumanDescription: "Delegate a focused task to a sub-agent and return its result.", AgentDefault: true, UserToggleable: true},
+	{Spec: GenerateImageToolSpec, HumanDescription: "Create an image from a written description.", AgentDefault: true, UserToggleable: true},
+	{Spec: CreateAgentJobToolSpec, HumanDescription: "Create a scheduled or recurring task for the assistant to run later.", AgentDefault: true, UserToggleable: true},
```
GoriVix: Here we might want at least some details like "jobs can be created in current or a new thread, accepts most natural language times (eg every tuesday at 2pm). Can override personality or running model"

internal/agent/tools/catalog.go
```diff
@@ -5,22 +5,23 @@ import "github.com/openai/openai-go/v3/responses"
...
+	{Spec: RecallToolSpec, HumanDescription: "Find relevant memories, files, summaries, and conversation history when more context is needed.", AgentDefault: true, UserToggleable: true},
```
GoriVix: Similar here, a little bit about use-cases would be great. Also probably a "Generally used with list to find a thing and then recall to get its contents"

GoriVix (commented):
For this to do anything, we also need to update the UI to replace the raw tool definitions with the human readable text, in the right sidebar

internal/agent/tools/catalog.go
dahmonson-ux: Fixed and verified against the implementation. The scratchpad human description now says it is this personality's working notes and that it persists across conversations using the same personality, rather than implying conversation-local scope. The agent-facing scratchpad prompt was not changed.

internal/agent/tools/catalog.go
dahmonson-ux: Fixed. I derived the list from the actual `list` implementation rather than the old copy. The human description now enumerates the supported kinds: models, personalities, skills, files, conversations, jobs, and MCP servers. The vague “and other resources” wording is gone.

internal/agent/tools/catalog.go
dahmonson-ux: Expanded this from the actual `create_agent_job` arguments. The human description now covers one-time/recurring natural-language scheduling, current-vs-new thread behavior, and optional model/skills overrides. One implementation detail I found while checking your example: this tool does not expose a direct personality-override argument, so I left personality override out rather than claiming a capability this tool cannot currently set.

internal/agent/tools/catalog.go
dahmonson-ux: Expanded this too. `find_context` now explains that it searches/retrieves memories, files, summaries, and conversation history, and calls out the common workflow of using `list` to find an item and then `find_context` to inspect/retrieve it. I kept `list` as a common workflow rather than a requirement because `find_context` can also be used directly.

dahmonson-ux:
@GoriVix review follow-up is in on `b480999`.

I traced the actual UI path before changing anything: `HumanDescription` in the catalog → `GetAvailableTools` → `/tools.description` → Angular `ToolService` → right-sidebar `ContextToolsTab` → `tool.description`. That showed the field was **not** being lost and the sidebar was already rendering the API description, so a second production frontend mapping was unnecessary.

What was actually missing was a defined display contract and some inaccurate/underspecified human copy. `GetAvailableTools` now trims/preferentially uses `HumanDescription` and falls back to the existing `Spec.Description` when the human field is absent/whitespace. That keeps external/runtime tools compatible and leaves the agent-facing tool prompts untouched.

I also audited the descriptions against code rather than the intended design:
- scratchpad now correctly says it persists across conversations for the same personality;
- `list` names the implemented resource kinds instead of “and other resources”;
- jobs now covers one-time/recurring natural-language scheduling, current/new threads, model override, and skills. One correction to the example in the review: `create_agent_job` does not currently expose a direct personality-override argument, so I did not claim that capability;
- `find_context` now gives the list → find_context use case without implying list is required;
- I also corrected `create_memory` scope wording and expanded `run_subagent` to its verified personality/model/skills options.

Regression coverage now includes the backend precedence/fallback rule, external-tool fallback, actual `/tools` serialization, and the right-sidebar component rendering the returned descriptions. On GitHub’s current-main merge ref `602537c` (`b480999` into current main `549fb71`), PR Validation is green including race-enabled Go tests/build/hermetic mock E2E; Frontend PR Validation is green including the exact context-tools sidebar spec (3/3), 1,125 passing frontend tests, and production build; local-LLM E2E is green. The separate long mock-backend E2E workflow is still running, although the hermetic mock E2E in PR Validation already passed.

BSD-main (14) final narrow evaluation: `eval-f60045af-1fcc-4865-8b89-85e32925c9c9`. Main mechanistic chain is `CERTIFIED_WITH_LIMITS`, structural reliability `0.704`, evidence mass `1.0` (coverage, not truth probability); the competing “UI/fallback still missing” chain was rejected. PPI is unavailable because there is no second surviving supported structure. I kept the BSD claim narrow to propagation + precedence/fallback + unchanged agent descriptions; the wording audit is separate and limited to the implementations inspected here.

I replied to and resolved the four actionable inline threads as well.

GoriVix (commented):
Thanks! Let me boot this up  real fast, make sure the UI works, then I will LGTM

GoriVix (approved):
LGTM

PR #72, opened 2026-08-30
Show estimated BYOK input cost in Context X-ray

### LINKED ISSUE #49: Context x-ray should show charged cost
### Area

Web app (web/app)

### Problem

We have all the billing data available, but do not expose that with the token information in the new right sidebar 'context xray' view.

### Proposed solution

We should add credit cost, likely near the top where we show the token widget.

### Alternatives considered

_No response_

### Additional context

_No response_

dahmonson-ux:
Fixes #49.

## Diagnosis

Context X-ray already has the right usage signal for a self-hosted/BYOK deployment: its `total_tokens` is reconciled to vendor-reported **input** usage when the provider returns it, and the snapshot already carries the exact provider + model used for the turn.

The earlier version of this PR added a hosted-style `charged_credits` field, but that made the open-source core depend conceptually on a billing authority it does not own. Self-hosters bring their own provider keys, so the useful OSS question is instead: **roughly what did the input context shown in this X-ray cost at the provider's standard API rate?**

## Implementation

- Remove the hosted `charged_credits` contract from this PR.
- Add one centralized frontend price catalog keyed by exact `provider/model` pairs.
- Calculate **Estimated input API cost** from the X-ray's existing vendor-reported input-token total.
- Show the reviewed per-million input rate and pricing review date in the X-ray footer.
- Use exact matches only. Unknown/custom/local models show no dollar estimate rather than inheriting a similar model's price.
- Keep pricing provenance with every catalog entry so maintainers can verify/update rates from the official provider source.

## Pricing freshness

The bundled rates were reviewed on **2026-08-30** against official provider pricing pages. The catalog currently covers seeded models whose standard text-input price could be verified confidently, including OpenAI GPT-5.4 / 5.1 / 5 / 5 Mini / 4.1 Mini / 4o / 4o Mini, Anthropic Claude Sonnet 4.6 / Haiku 4.5, and Google Gemini 3.5 Flash.

A seeded or user-added model without a confidently verified exact rate is deliberately left unpriced until its rate is added. That is preferable to showing a plausible-looking but unsupported number.

## Important boundary

This is an estimate of the **input context shown by Context X-ray**, not a reconstruction of the provider invoice. It intentionally does not claim to include output tokens, cached-input discounts, tool/search fees, batch or priority tiers, regional uplifts, free-tier effects, promotions, or account-specific pricing. Those require billing/usage information that this X-ray does not currently persist.

That boundary is visible in the UI rather than hidden in implementation details.

## Test-first evidence

The revised regression first required an estimated provider cost and failed against the prior hosted-credit implementation. Source tracing then showed that Context X-ray has authoritative input usage but does not persist authoritative output usage, so the regression was narrowed to the quantity this view can support without inventing data: standard input API cost.

Coverage now verifies:
- a known exact provider/model calculates the expected input cost;
- reviewed OpenAI, Anthropic, and Gemini entries resolve correctly;
- unknown/local/custom models produce no estimate;
- the Context X-ray renders the estimate beside the token gauge.

## Epistemic boundary

A green test suite establishes the encoded lookup/calculation/display behavior. It does not establish that the estimate equals a user's provider invoice, and the UI explicitly says why. Provider pricing can change, so the review date and source-backed centralized catalog are part of the feature rather than an implied guarantee of permanent accuracy.

Current-head evaluation: `eval-7b853c5c-1c87-4400-9d16-25e06af2bda2`.

The exact-rate input-cost mechanism remains technically supported and the current Frontend + local-LLM + mock E2E workflows are green. The architecture remains genuinely contested, however: the maintainer requested a pluggable cost-widget seam so OSS can supply a rough estimate while the closed-source path can supply billed cost, and that seam is not present in the current diff. BSD therefore keeps implementation-correct and partial-correctness at equal comparison weight rather than treating green CI as resolution of that design requirement.


GoriVix:
Ah this is a good point. Ok let's make the widget pluggable and then for the OSS one we can plug in an impl that gives a rough cost and then for the closed source one, we can give billed cost. 

Hmm, we PROBABLY want to replace the billed costs with data on the model objects in the backend, but we can do that a bit later, we still need to wire in the model-management UI for OSS so that would be a good time to move it to the model object (and then honestly we can reuse the existing credits fields to store it!)

@LiminalDevving  as this is relevant to the OSS tidying

web/app/src/app/features/chat/helpers/api-pricing.helpers.ts
```diff
@@ -0,0 +1,89 @@
...
+  pricingCheckedAt: string;
+}
+
+const PRICING_CHECKED_AT = '2026-08-30';
+
+// Standard pay-as-you-go text-input rates reviewed against official provider pricing.
+// Exact provider/model keys are deliberate: custom and unknown models must not inherit a
+// nearby model's price. Cached-input discounts, batch/priority tiers, regional uplifts,
+// tool-call fees, and promotional account-specific pricing are outside this estimate.
+const INPUT_TOKEN_PRICES: Record<string, InputTokenPrice> = {
```
GoriVix, line 22: This is fine-ish for now, but @LiminalDevving When we add model control to OSS, we should make these fields on the model object (and use the extensions so it can be OSS only, as for hosted we are gonna replace this with credit costs to simplify for users)

GoriVix (commented):
Image

Tidied up a little bit, shorter text and made it pluggable for billing / credit system!

Note to self will need to do equivalent internal work when this one gets merged upstream.

PR #73, opened 2026-08-30
Add emoji shortcode expansion in chat composer

### LINKED ISSUE #50: Add emoji via :<emoji> : syntax
### Area

API / backend (Go)

### Problem

In chat, you have to open the emoji picker always to add an emoji. In most apps, you can type eg 🦊  \:fox\: to get an emoji

### Proposed solution

IDK how this works elsewhere but seems like a reasonably standard type library or other such widget we can add.

### Alternatives considered

_No response_

### Additional context

_No response_

dahmonson-ux:
Fixes #50.

## Diagnosis

The issue is labeled backend, but the current architecture points to the chat editor as the correct ownership boundary. The web app already ships `@ctrl/ngx-emoji-mart`, the composer already owns native emoji insertion, and rewriting message text server-side would make transport/persistence mutate user-authored content.

The first red run confirmed the missing behavior cleanly: Frontend PR Validation and local-LLM E2E passed, while mock E2E failed only the new recognized-shortcode case. The companion regression proving unknown shortcode text stays literal already passed.

## Test-first evidence

The first commit was test-only and added two browser regressions:

- typing `Hello :fox:` must produce `Hello 🦊` in the composer;
- typing an unknown shortcode must remain unchanged.

At head `eaa886b`, the recognized case failed on both initial run and retry with literal `Hello :fox:`. The shard otherwise passed 26 tests (1 skipped), giving a narrow red signal.

## Implementation

- Reuse the installed Emoji Mart `EmojiSearch` catalog; no second emoji dependency or hard-coded emoji table.
- Detect only a completed `:shortcode:` immediately before the caret and only when it begins at a safe editor boundary.
- Prefer exact emoji IDs/short names. For convenience aliases such as `:fox:` → `fox_face`, accept a word-level search match only when it is unambiguous.
- Reject ambiguous fuzzy matches and leave unknown/URL-like colon text literal.
- Replace the token before Angular's controlled-input handler consumes the event, so the existing draft/autosave/send path receives the expanded Unicode value and the caret remains after the inserted emoji.
- Keep the backend and persisted-message transport untouched.

Additional helper-level regressions cover exact-vs-fuzzy precedence, the `fox` alias, ambiguity rejection, unknown queries, caret-local matching, and URL-ish text.

## Earlier BSD boundary

An earlier red-stage BSD pass kept evidence provenance/PPI unavailable because no host-bound verifier receipts or contrastive probes had been supplied at that stage. That pass was useful as a scope check: the observed red state, existing composer ownership, and installed search API supported the editor-side patch, while fuzzy-match false positives remained the explicit counterexample the implementation had to guard against.

CI on the implementation head remains the acceptance gate; this description does not treat BSD or the local source trace as proof that the patch passes.

Current-head evaluation: `eval-c329714e-c6ba-4c5b-bac4-05c879097beb`.

The current-head composer implementation remains supported for the represented behavior: safe-boundary shortcode detection, exact-name precedence, unique whole-word alias fallback, literal unknown/ambiguous text, and the current Frontend + local-LLM + mock E2E workflows are green. The partial frame is net-refuted for that bounded slice, but broader Emoji Mart catalog/property coverage remains outside the represented evidence.

### DIFF
--- a/web/app/e2e/poms/chat.page.ts
+++ b/web/app/e2e/poms/chat.page.ts
@@ -60,9 +60,20 @@ export class ChatPage {
     this.skillOptionList = this.skillPickerDialog.getByRole('listbox', { name: 'Available skills' });
     this.skillPickerStatus = this.skillPickerDialog.locator('.composer__skill-status');
     this.pendingSkills = this.page.getByLabel('Skills to send with this message');
+    this.emojiAutocomplete = this.page.getByRole('listbox', { name: 'Emoji suggestions' });
   }
 
   readonly composerInput: Locator;
+
+  // --- emoji shortcode autocomplete -----------------------------------------
+
+  /** The `:shortcode` suggestion popup (`role="listbox"` "Emoji suggestions"). */
+  readonly emojiAutocomplete: Locator;
+
+  /** A single emoji suggestion row, matched by its `:short_name:` label. */
+  emojiSuggestion(colons: string): Locator {
+    return this.emojiAutocomplete.getByRole('option', { name: colons });
+  }
 
   /**
    * Header button showing the active personality's name (and the matching
--- a/web/app/e2e/tests/functional/chat/emoji-shortcodes.spec.ts
+++ b/web/app/e2e/tests/functional/chat/emoji-shortcodes.spec.ts
@@ -1,0 +1,23 @@
+import { test, expect } from '../../../fixtures';
+
+test('expands a completed emoji shortcode in the composer', async ({ chatPage, seed, userWithPersonality }) => {
+  const thread = await seed.thread(undefined, {
+    personalityId: userWithPersonality.personality.id,
+  });
+  await chatPage.navigateTo(thread.id as string);
+
+  await chatPage.composerInput.pressSequentially('Hello :fox:');
+
+  await expect(chatPage.composerInput).toHaveValue('Hello 🦊');
+});
+
+test('leaves unknown emoji shortcodes untouched', async ({ chatPage, seed, userWithPersonality }) => {
+  const thread = await seed.thread(undefined, {
+    personalityId: userWithPersonality.personality.id,
+  });
+  await chatPage.navigateTo(thread.id as string);
+
+  await chatPage.composerInput.pressSequentially('Keep :not_a_real_emoji: literal');
+
+  await expect(chatPage.composerInput).toHaveValue('Keep :not_a_real_emoji: literal');
+});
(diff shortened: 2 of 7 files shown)


GoriVix:
Updating to be more slack-like instead of silent replace. 

Compass review:

Files under review: 7

Round 1: Executing tool call: git_diff [{"context_lines":3,"file_path":"","target":"staged"}]
📋 Code Review Results: 5 issue(s) found

ℹ️  MINOR SUGGESTIONS
================================================================================

1. web/app/src/app/features/chat/components/chat-composer/chat-composer.component.ts
   Lines 1427-1308
      In `updateEmojiAutocomplete`, a thrown error from `emojiSearch.search`
   would currently bubble up and potentially break input handling. While
   unlikely, it might be safer to guard this call with a small try/catch and
   treat failures as `[]`, so a misconfigured emoji index doesn’t break the
   composer’s typing experience.

2. web/app/src/app/features/chat/components/chat-composer/chat-composer.component.ts
   Lines 1408-1317
      `updateEmojiAutocomplete` is doing several things (caret validation,
   pattern matching, querying, state and UI control). For readability and
   future changes, consider extracting small pure helpers (e.g.
   `getActiveEmojiShortcode(textarea)` or `computeEmojiSuggestions(active)`)
   so this method remains focused on orchestration rather than detailed
   logic.

3. web/app/src/app/features/chat/components/emoji-autocomplete-menu/emoji-autocomplete-menu.component.ts
   Lines 73-214
      `EmojiAutocompleteMenuComponent` duplicates some keyboard handling
   (Escape, ArrowUp/Down) that is also partially implemented in the
   parent’s `onKeydown`. This split of responsibilities is subtle and could
   become fragile. Consider centralizing key handling responsibilities
   clearly (e.g. menu owns all navigation and close keys, parent only
   delegates raw events) and documenting that contract to avoid regressions
   when either side is changed.

4. web/app/src/app/features/chat/helpers/emoji-shortcode.helpers.ts
   Lines 64-149
      In `searchEmojiShortcodes`, `emojiSearch.search` is called on each input
   event with `normalized` and `limit`. This is expected, but depending on
   the cost of the underlying search, you may want to consider basic
   debouncing or short-circuiting when the query hasn’t changed since the
   last call to avoid unnecessary work while the user is rapidly typing.

💡 NITPICKS
================================================================================

1. web/app/src/app/features/chat/components/emoji-autocomplete-menu/emoji-autocomplete-menu.component.ts
   Lines 89-200
      The `effect` in `EmojiAutocompleteMenuComponent` that resets
   `selectedIndex` on every `suggestions()` change is correct but a bit
   implicit. A short comment describing that this is to ensure the highlight
   resets when the query changes (and that suggestions are read-only input)
   will help future readers understand why a reactive effect is used instead
   of, for example, resetting in an `@Input` setter.

================================================================================
✅ No critical issues found

================================================================================
✅ Code review passed gating check
image

PR #75, opened 2026-08-31
Restore Run Now behavior on current job detail

dahmonson-ux:
Follow-up to merged PR #61 / issue #37.

## Diagnosis

#61 implemented the Run Now breadcrumb in the old `AgentJobDetailComponent`. The current `/agent-jobs/:id` route now renders `JobDetailPageComponent`, whose pre-fix `runNow()` only triggered the API and restarted polling. The merged behavior was therefore lost when the job-detail UI was replaced.

The current page also rendered Run Now for terminal jobs even though the backend only accepts active/paused jobs.

## Fix

- Restore the successful Run Now breadcrumb on the component the router actually uses.
- When an active/paused job already has a chat, successful Run Now selects that chat and navigates to `/chat`.
- Jobs without an existing chat keep the current polling behavior.
- Hide and method-guard Run Now for non-runnable terminal statuses.

## Regression coverage

- successful Run Now on a job with `chat_id` must call the API, select that chat, and navigate to `/chat`;
- terminal job detail must not expose Run Now.

This intentionally does not add live execution-state UI or change scheduler semantics.

## Validation

Head `d95217c`:

- Frontend PR Validation: **passed** (type-checking, unit tests, coverage, production build).
- Separate long mock-backend E2E: queued at the time of this update and not represented as passed.

## BSD-main (14) final narrow evaluation

Evaluation `eval-9e8e4690-0787-4079-863b-41741d43f0ab`.

Evaluated claim: the active routed job-detail component restores #61's associated-chat breadcrumb for successful Run Now, matches backend active/paused eligibility, preserves no-chat polling, and does not redesign scheduler execution.

- main mechanistic chain: `CERTIFIED_WITH_LIMITS`
- structural reasoning reliability: `0.75` (`MEASURED`; not truth probability)
- competing “still only trigger/poll” chain: `REJECTED`

BSD limitations: source and regression tests are one dependent evidence group; the separate mock-browser E2E was still pending; no manual live-browser click was captured; GitHub evidence is caller-supplied/unverified in the local BSD runtime.

### DIFF
--- a/web/app/src/app/features/agent-job/job-detail-page.component.html
+++ b/web/app/src/app/features/agent-job/job-detail-page.component.html
@@ -18,14 +18,16 @@
             Open chat
           </button>
         }
-        <button
-          type="button"
-          class="rounded-lg border border-border-base px-3 py-2 text-xs text-text-secondary"
-          [disabled]="runningNow()"
-          [attr.aria-disabled]="runningNow()"
-          (click)="runNow()">
-          {{ runningNow() ? 'Running…' : 'Run now' }}
-        </button>
+        @if (canRunNow()) {
+          <button
+            type="button"
+            class="rounded-lg border border-border-base px-3 py-2 text-xs text-text-secondary"
+            [disabled]="runningNow()"
+            [attr.aria-disabled]="runningNow()"
+            (click)="runNow()">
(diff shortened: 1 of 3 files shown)


GoriVix (approved):
LGTM -- we need to fix the UI in chat for running jobs, but that is separate.

PR #65, opened 2026-08-29
Add personality prompt change history

dahmonson-ux:
Fixes #42.

## Diagnosis

`UpdatePersonality` previously overwrote `system_prompt` without recording a durable before/after history. The existing `scratchpad_history` JSON list is a bounded text-history mechanism, while #42 asks for change-log/audit semantics with reversion and asks that those entries appear in the existing compaction/change-log UI.

## Implementation

- Add a dedicated append-only personality-prompt change record scoped to the owning user/personality.
- Record ordinary edits as old → new prompt transitions.
- Reverting restores the captured prior prompt and appends a new `revert` audit event instead of mutating history.
- Add owner-scoped list and revert endpoints.
- Route personality PUT through the history-aware update path.
- Surface personality prompt audit entries on the existing Compaction log page, alongside checkpoint history.
- Show personality, action, timestamp, before/after prompt, and a real `Restore previous` action.
- Respect the existing personality filter when loading prompt audit history.

## Test-first / regression evidence

The first regression required explicit datastore seams for listing prompt changes and reverting one and went red before those seams existed.

Behavior coverage now exercises the actual datastore and UI paths:

- history is owner-scoped and newest-first;
- reverting restores the captured old prompt;
- a revert appends a new audit event while preserving the original change;
- another user cannot revert someone else's change;
- repeated restoration is idempotent rather than duplicating audit history;
- the Compaction log renders the prompt audit entry with before/after content;
- the UI restore action calls the revert endpoint and refreshes the audit list;
- personality filtering narrows the prompt-history request.

## BSD boundary

BSD was used to constrain the design and keep the audit record append-only; it is not treated as independent proof of correctness.

## Validation

Exact current head `2605349` passes PR Validation, Frontend PR Validation, local-LLM E2E, and mock-backend E2E.

## BSD rerun — coding profile / current engine

[...]

PR #76, opened 2026-08-31
Make personality prompt history collapsible

dahmonson-ux:
Follow-up to merged PR #65.

## Diagnosis

#65 added personality prompt audit history above the existing compaction/checkpoint event feed. The prompt-history cards render fully expanded by default, so a nontrivial prompt history can push the older checkpoint feed far down the page.

## Fix

- Keep the Personality prompt changes section visible as a compact header/description.
- Collapse its detailed audit cards by default.
- Add an accessible toggle using `aria-expanded`.
- Preserve the existing audit loading, filtering, before/after diff, and restore behavior when expanded.

## Regression coverage

The component regression asserts that prompt-history details are hidden by default while the checkpoint feed remains visible, then verifies the same audit details and restore action appear after expansion.

This is a presentation-only follow-up. It does not change the append-only audit data model or revert semantics.

## Validation

Head `f680bbb`:

- Frontend PR Validation: **passed** (type-checking, unit tests, coverage, production build).
- Separate long mock-backend E2E: still running at the time of this update and not represented as passed.

## BSD-main (14) final narrow evaluation

Evaluation `eval-62a0b044-07b7-49a1-8da0-61ec33cbdd53`.

Evaluated claim: prompt-history details default to collapsed so they do not crowd the older checkpoint feed, remain discoverable through an accessible toggle, and preserve the existing audit/revert content when expanded.

- main mechanistic chain: `CERTIFIED_WITH_LIMITS`
- structural reasoning reliability: `0.75` (`MEASURED`; not truth probability)
- competing “still crowded by default” chain: `REJECTED`

BSD limitations: source and component regression are one dependent evidence group; the long mock-browser E2E was still running; no visual screenshot was captured; GitHub evidence is caller-supplied/unverified in the local BSD runtime.


GoriVix (approved):
LGTM - much better

PR #81, opened 2026-09-01
feat(agent): push-notification seam (internal/pushnotify)

GoriVix:
## Push-notification seam (`internal/pushnotify`)

Adds a supported extension seam so a completed assistant reply can be delivered out-of-band (push, email, …), following the exact shape of the existing `metering.Meter` seam. This lets a private/operator package deliver notifications without patching the core.

### What's here
- **`internal/pushnotify`** — `Notifier` interface + a plain `Event{UserID, ChatID, MessageID, Body}` + `NoopNotifier` + `var New func(ds, logger) Notifier` (nil unless a package sets it in `init()`).
- **Agent wiring** — `AgentConfig.PushNotifier`, `Agent.pushNotifier`, and `NewAgent` falls back to `NoopNotifier` when none is supplied.
- **Server wiring** — `if pushnotify.New != nil { agentCfg.PushNotifier = pushnotify.New(dataStore, logger) }`, alongside the meter/featuregate/userhooks seams.
- **Hook** — invoked at the agent-job completion tail in `agentjob_run.go`, guarded `callPath == telemetry.CallPathAgentJob`, so scheduled/autonomous and webhook background-mode replies notify while interactive replies do not.

### Notes
- **No new dependencies.**
- **No behavior change for the OSS build**: with no notifier linked (the default), `New` is nil and completed replies notify nothing.

### DIFF
--- a/internal/agent/agentjob_run.go
+++ b/internal/agent/agentjob_run.go
@@ -11,6 +11,7 @@ import (
 	"github.com/theimaginaryfoundation/what-iff/internal/metering"
 	"github.com/theimaginaryfoundation/what-iff/internal/middleware"
 	"github.com/theimaginaryfoundation/what-iff/internal/models"
+	"github.com/theimaginaryfoundation/what-iff/internal/pushnotify"
 	"github.com/theimaginaryfoundation/what-iff/internal/telemetry"
 
 	"github.com/google/uuid"
@@ -319,6 +320,23 @@ func (a *Agent) handleEphemeralPrompt(
 	}
 	if err := a.advanceChatJobStatus(ctx, trackingJob, models.JobStatusComplete); err != nil {
 		return nil, fmt.Errorf("finalize agent job: %w", err)
+	}
+
+	// Push a notification for this completed reply. Gated to the agent-job path
+	// (callPath), which is what both scheduled/autonomous jobs and webhook
+	// background-mode invocations run under — the replies that land after the
+	// user has closed the app. Interactive user chat completes elsewhere
+	// (runUserChatPostInferencePhases) and is deliberately not notified.
+	// Fire-and-forget on the lifecycle context so the send outlives this job's
+	// request context but still stops on process shutdown; NoopNotifier makes it
+	// a no-op in builds without a push implementation.
+	if callPath == telemetry.CallPathAgentJob {
+		a.pushNotifier.Notify(a.lifecycleCtx, pushnotify.Event{
+			UserID:    userID,
+			ChatID:    agentMessage.ChatID,
+			MessageID: agentMessage.ID,
+			Body:      agentMessage.Message,
+		})
 	}
 
 	return agentMessage, nil
--- a/internal/agent/agentjob_run.go
+++ b/internal/agent/agentjob_run.go
@@ -322,21 +322,33 @@ func (a *Agent) handleEphemeralPrompt(
 		return nil, fmt.Errorf("finalize agent job: %w", err)
 	}
 
-	// Push a notification for this completed reply. Gated to the agent-job path
-	// (callPath), which is what both scheduled/autonomous jobs and webhook
-	// background-mode invocations run under — the replies that land after the
-	// user has closed the app. Interactive user chat completes elsewhere
-	// (runUserChatPostInferencePhases) and is deliberately not notified.
-	// Fire-and-forget on the lifecycle context so the send outlives this job's
-	// request context but still stops on process shutdown; NoopNotifier makes it
-	// a no-op in builds without a push implementation.
-	if callPath == telemetry.CallPathAgentJob {
-		a.pushNotifier.Notify(a.lifecycleCtx, pushnotify.Event{
+	// Notify the user of this completed reply. Gated to the agent-job path
+	// (callPath) — scheduled/autonomous jobs and webhook background-mode
+	// invocations, the replies that land after the app is closed; interactive
+	// chat completes elsewhere (runUserChatPostInferencePhases) and is not
+	// notified. Only pointers are passed; the client fetches the content.
+	//
+	// Runs on a detached goroutine with panic recovery so a slow or faulty
+	// notifier implementation can neither stall nor crash this job goroutine.
+	// a.lifecycleCtx outlives the request but is cancelled on shutdown. Skipped
+	// entirely when no push implementation is wired (a.pushEnabled == false).
+	if a.pushEnabled && callPath == telemetry.CallPathAgentJob {
+		ev := pushnotify.Event{
+			Kind:      pushnotify.EventKindAgentReply,
 			UserID:    userID,
 			ChatID:    agentMessage.ChatID,
 			MessageID: agentMessage.ID,
-			Body:      agentMessage.Message,
-		})
+		}
+		go func() {
+			defer func() {
+				if r := recover(); r != nil {
+					a.logger.Error("push notifier panicked",
+						zap.Any("recover", r),
+						zap.ByteString("stack", debug.Stack()))
+				}
+			}()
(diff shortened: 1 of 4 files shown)


csotherden:
⚠️  MAJOR ISSUES
================================================================================

1. internal/agent/agentjob_run.go
   Lines 319-52
      handleEphemeralPrompt unconditionally invokes a.pushNotifier.Notify on the
   lifecycle context, and the Notifier contract says implementations must not
   block and must swallow their own errors. If an implementation is
   misconfigured (e.g., a nil New assigned at runtime, or a Notify that
   panics), failures would happen out-of-band and might be hard to diagnose.
   Consider adding a small wrapper Notifier inside the agent that recovers
   from panics in Notify and logs them, so push failures never crash job
   workers and are at least visible in logs.
   
Additionally, should we consider making this seam more generic? Just like a general event emitter that multiple things could subscribe to? Then you could register a listener for push notifications or _whatever_ not just have a push notification specific seam.

GoriVix:
+1 to more generic and the error handling. Yeah, I dont see any reason not to have events at the end of *every* inference or even job state change (not that I'd go that far now...) and it really is just a generic event listener. Honestly has a ton of use cases

LiminalDevving:
Had a look at this. Seam shape looks right to us: same five pieces as `metering`, in the same order. The `callPath` guard is also doing real work, since `HandleEphemeralPromptSync` passes `CallPathUserChat` into the same function, so without it interactive replies would notify.

Two things to add to @csotherden's notes rather than repeat them.

**`Event.Body` carries more than the consumer needs.** It is the full assistant reply, unbounded, and the doc puts the truncation obligation on the implementation. But the implementation already gets `*datastore.Datastore` plus `UserID` and `MessageID`, and `GetChatMessage(ctx, userID, id)` exists, so it could fetch and truncate on its own.

Our reason for changing it: this is the first seam to move message content across the boundary. `metering.Usage` carries a user ID, action type and token counts. `userhooks` carries a user record. This carries what the assistant said, in full, into a struct an implementation might log wholesale. A bounded `BodyPreview`, truncated in the core to a couple of hundred characters, keeps the single read and makes the truncation the core's decision instead of a documented duty.

**Nothing pins the `callPath` branch,** which is effectively the whole feature. Codecov reports 0% patch coverage on `agentjob_run.go` for the same reason. A fourth caller passing `CallPathAgentJob`, or that condition getting flipped, would notify on every interactive reply and no test would catch it. One case each way would cover it: `HandleAgentJobPrompt` notifies, `HandleEphemeralPromptSync` does not.

**On the generic event emitter:** we would lean against, and the `Body` point above is why. The four existing seams are narrow and typed, and that narrowness is what makes the boundary auditable. You read `Usage` and you know exactly what crosses. A generic payload has to satisfy every future subscriber, so it grows toward including everything. Even this purpose-built seam already ships a field its consumer does not need. `plugins` is also already the general extension point for private packages, so an emitter would sit alongside it as a fifth pattern rather than replacing anything.

One small thing: the note above says `Notify` is invoked unconditionally, but it is guarded on `callPath == telemetry.CallPathAgentJob`. Does not change the panic point, just flagging so nobody adds a guard that is already there.

csotherden:
> Had a look at this. Seam shape looks right to us: same five pieces as `metering`, in the same order. The `callPath` guard is also doing real work, since `HandleEphemeralPromptSync` passes `CallPathUserChat` into the same function, so without it interactive replies would notify.
> 
> Two things to add to @csotherden's notes rather than repeat them.
> 
> **`Event.Body` carries more than the consumer needs.** It is the full assistant reply, unbounded, and the doc puts the truncation obligation on the implementation. But the implementation already gets `*datastore.Datastore` plus `UserID` and `MessageID`, and `GetChatMessage(ctx, userID, id)` exists, so it could fetch and truncate on its own.
> 
> Our reason for changing it: this is the first seam to move message content across the boundary. `metering.Usage` carries a user ID, action type and token counts. `userhooks` carries a user record. This carries what the assistant said, in full, into a struct an implementation might log wholesale. A bounded `BodyPreview`, truncated in the core to a couple of hundred characters, keeps the single read and makes the truncation the core's decision instead of a documented duty.
> 
> **Nothing pins the `callPath` branch,** which is effectively the whole feature. Codecov reports 0% patch coverage on `agentjob_run.go` for the same reason. A fourth caller passing `CallPathAgentJob`, or that condition getting flipped, would notify on every interactive reply and no test would catch it. One case each way would cover it: `HandleAgentJobPrompt` notifies, `HandleEphemeralPromptSync` does not.
> 
> **On the generic event emitter:** we would lean against, and the `Body` point above is why. The four existing seams are narrow and typed, and that narrowness is what makes the boundary auditable. You read `Usage` and you know exactly what crosses. A generic payload has to satisfy every future subscriber, so it grows toward including everything. Even this purpose-built seam already ships a field its consumer does not need. `plugins` is also already the general extension point for private packages, so an emitter would sit alongside it as a fifth pattern rather than replacing anything.
> 
> One small thing: the note above says `Notify` is invoked unconditionally, but it is guarded on `callPath == telemetry.CallPathAgentJob`. Does not change the panic point, just flagging so nobody adds a guard that is already there.

I guess my thought around this would be that the generic event seam would be this action completed, here's the user ID, here's the status, and here's the ID of the thing generated. Then we don't pass around everything everywhere (like full response message bodies) and consumers are responsible for enriching whatever raw details they need. Which, maybe I'm trying to make it too Wordpress-y, but it just feels like we're already building in more stuff that gets invoked on these message lifecycle points and now it's making me wonder if the right way to make that extensible instead of continuing to add more specific seams is just emitting events at those lifecycle hooks. I'm not sure though if that's making extensibility we don't need and where to draw the line between this is core message lifecycle processing and this is an extension lives, but it feels like this is the type of spot we're just going to keep adding more and more hooks into.

Just thinking out loud on this. If the panic possibility gets cleaned up I'm not opposed to just proceeding as is under YAGNI and re-evaluate some of the message processing lifecycle stuff later.

GoriVix:
@csotherden 
Ok, I will fix the panic and let's go with YAGNI on it for now -- I dont think we will get the interface right at the moment anyways and it would be a shame to block real notification use-cases on this for weeks while we debate packet structure. I'll try to generify it a little bit then make a note to  circle back!

GoriVix:
Thanks for the review — pushed `bc6beb0` addressing all three:

1. **Pointers, not content.** `Event` no longer carries the reply `Body` — just `UserID`/`ChatID`/`MessageID`. The recipient fetches the content via the API, so message text never transits the notifier (nor, in the FCM impl, a third-party push service). Nice privacy win too.
2. **A bad impl can't take down the loop.** The core now invokes `Notify` on a **detached goroutine with `recover()`**, gated by a `pushEnabled` flag. Doing it core-side (rather than trusting the impl) is deliberate: a panic in an impl's own goroutine would otherwise crash the process. So a slow *or* panicking notifier can neither stall nor crash the agent-job goroutine.
3. **A bit more generic (YAGNI-bounded).** Added `Event.Kind` (`EventKindAgentReply`) so the same seam can carry other user-facing actions later — without a bus/registry or widening the interface. Stopped there.

The private FCM implementation was updated to match (`Notify` is now synchronous since the core owns the goroutine; sends a generic body + pointer data payload).

LiminalDevving (approved):
LGTM. `bc6beb0` covers all three.

Dropping `Body` outright is better than the truncated preview we suggested — content never transits the notifier or a third-party push service at all, and the exported surface got smaller rather than larger, which is the part that's hard to walk back in a public repo. Owning the goroutine and the `recover()` in the core is the right call for the reason you gave: a panic in an impl's own goroutine takes the process down, so the core has to own the goroutine to own the recover.

csotherden (approved):

PR #87, opened 2026-09-04
fix(agent): fail empty model turns instead of saving blank messages

LiminalDevving:
## Summary

- A provider call can return cleanly and still carry no assistant text — a stream that closes after `message_start` without content blocks, a response truncated before any text, or content in a block shape the extractor does not match. Nothing checked for that, so the empty result was saved as an ordinary assistant message: the turn looked answered while being blank, no error surfaced, no retry was offered, and because `AppendHistoryTurn` skips empty content the row then vanished from the rebuilt context on the next turn. `assertGenerationProducedOutput` now fails the turn instead, after tool-call metrics and the web-search count so billable work stays counted. Image-only turns are still allowed.
- `GenerateResponse` carries `StopReason` (Anthropic `Message.StopReason`; Responses `IncompleteDetails.Reason` falling back to `Status`), logged with `output_tokens` — together they distinguish "text was generated and dropped in extraction" from "nothing came back". `BuildClaudeParamsWithMaxTokens` gives the Anthropic path the per-call output cap the Responses path already had.
- Memory context no longer duplicates itself across turns. `FormatMemoryForContext` appends metadata computed at render time, so a memory persisted on an earlier turn and the same memory rendered now are different strings; keying dedupe on that string added a near-duplicate per turn — same content, same `stored_at`, a different `age_days` on each. Memory items are now keyed on metadata-stripped normalised content (the identity every other dedupe site in the package already uses), and a duplicate keeps its position but takes the newer rendering.

## Test plan

- [x] `make pre-commit`
- [x] Manually verified: full local CI replay via `make ci-local --e2e` — 22/22 checks passed, 0 skipped (go incl. `test-ci` with race + coverage, mock-e2e, frontend, Playwright: 104 browser + 51 API, 0 failed, 0 flaky). Confirmed both new tests fail without their fix rather than passing vacuously — removing the guard call panics `TestRunGeneration_FailsTurnOnEmptyModelResponse`, and reverting the dedupe identity leaves 2 entries where the test expects 1.

## Checklist

- [ ] Regenerated the e2e SDK (`cd web/app && npm run sdk:generate`) if `openapi.yaml` changed


GoriVix (approved):
LGTM

We probably need to do more work on failures / partial failures soon, as it gets annoying and confusing for users. but this seems like an improvement as-is

PR #94, opened 2026-09-09
fix: Fail the import job when no conversation could be persisted

unlikelyzero:
## Summary

Two things, the second built on the first.

**The fix.** `ImportChats` returns an error only for a failure that aborts the whole run. A conversation that fails on its own is appended to `ImportResult.Errors` and the loop carries on, which is right for one bad thread among many. Nothing distinguished that from the case where it happened to *every* conversation: the job was still marked `complete`, the terminal progress payload still said `imported: 0`, and the importer rendered "Imported 0 threads" with no error anywhere the user could see. The only record that the entire import had failed was a `Warn` line on the server.

A run that imported nothing **and** skipped nothing is now marked `failed`. Zero on both counters means no conversation reached the database and none was recognised as a duplicate, so there is no reading of it that constitutes success. The two neighbouring zero-import cases are unchanged and now covered by tests: re-importing an export already in the account skips everything and still completes, because that is the no-op the UI reports as "No new threads", and an empty archive still completes.

**The coverage.** The rest of this PR is the first slice of end-to-end coverage for chat and memory import/export — 19 tests across four spec files, plus the harness they need. It starts here because the fix above is the sort of thing only this layer catches: the `(owner_id, import_hash)` unique index that makes a re-import a no-op exists only in Postgres, and both export endpoints stream, so they commit to a 200 and the ZIP headers before reading a single row. A handler test asserts on the bytes it handed a recorder, not the bytes that survived the wire.

| Spec | Layer | Tests |
| --- | --- | --- |
| `e2e/api-tests/chat-import-job.spec.ts` | API | 6 |
| `e2e/api-tests/chat-export.spec.ts` | API | 3 |
| `e2e/api-tests/memory-portability.spec.ts` | API | 6 |
| `e2e/tests/functional/chat/chat-export.spec.ts` | Browser | 4 |

Supporting harness: `readZipEntries`/`parseJsonl`/`readDownloadedZip` in `e2e/helpers/zip.ts`, memory-archive fixture builders, `waitForJobTerminal` in the api-test fixtures (`waitForJobComplete` throws on `failed`, which these specs must observe), and import/export/job-mutation helpers in the e2e SDK. `listChats` gains query params and an archived pager, because import creates threads archived and the endpoint returns only active ones by default.

## Defects this pins

Four tests run under `test.fail`. Each was verified against a running backend rather than assumed, and each turns red the moment the behaviour is fixed, which is the signal that the expectation is stale.

- **`PUT /job/{id}/status` answers 500 for an unrecognised status.** `openapi.yaml` documents 400 for invalid request data. `datastore.ErrInvalidJobStatus` exists and the handler maps it to 400, but the real datastore never returns it, so the value reaches ent's enum validator and comes back as a generic server fault.
- **`POST /memory/import` answers 500 for a misnamed entry.** An entry called `personality-backup.json` matches the personality-file prefix, so the importer parses `backup` as a UUID and fails. A user who renamed `user.json` is indistinguishable from an outage, in the response and in the logs.
- **Non-ASCII thread names download mojibaked.** A thread seeded as `Café ☕ 日本語` arrives as `CafÃ© â__ æ_¥æ_¬èª_`. `Content-Disposition` carries a bare `filename=`, which RFC 6266 reads as latin-1; the fix is a `filename*=` parameter in UTF-8 that the client then prefers.
- **A failed export reports "An error occurred".** The response is read as a blob whatever the status, so the 404's JSON envelope never reaches the error formatter and the server's "Chat not found" never reaches the user.

None of these are regressions from this PR, and none are fixed here — splitting each into its own change keeps this diff reviewable.

## Notes for review

- `test.fail` rather than `test.fixme` for the four above. The suite's existing `fixme` cases are skipped because they cannot run at all; these run and demonstrate the defect correctly, so keeping them executing means the assertion cannot rot.
- One plan assumption was wrong and the test now asserts reality: an empty archive completes at the handler layer, but over HTTP an empty JSON array is refused at upload with 400 before a job exists. The two cannot both be reached from a browser.
- Four `bookmarks.spec.ts` API tests fail on this branch with a 404. They fail identically on `main` and are untouched by this diff.

## Test plan

- [x] `make pre-commit`
- [x] `npm run lint:e2e` clean
- [x] `npx playwright test --config=e2e/playwright.config.mock-llm.api.ts` — 62 passed, 4 pre-existing bookmark failures
- [x] `npx playwright test --config=e2e/playwright.config.mock-llm.ts chat-export` — 5 passed, 7 per-project skips
- [x] Manually verified: the new Go test fails against the unfixed handler, reporting status `complete` where `failed` is expected, and passes with the fix.
- [x] Manually verified: the all-duplicates test passes both before and after, so the change narrows the success case rather than inverting it.
- [x] `importStore` now records the message sent with the terminal status, so the test asserts what the client is actually told rather than only the status value.

## Checklist

## Merge order

This is one of three parallel fixes to the chat importer, and GitHub reports all three mergeable only because it compares each against `main` in isolation. Verified by merging them in sequence locally: once #92 and #93 land, this branch conflicts with them in `internal/handlers/chat/_PACKAGE_SUMMARY.md`, where all three extend the same import bullet. It is a text collision in one documentation line, not a disagreement about behaviour, so it needs a rebase rather than an auto-merge.

Suggested order: #92, then #93, then this one. With all three merged, the combined tree builds and `internal/handlers/chat`, `internal/models` and `internal/chatimport` all pass.


GoriVix:
Thanks @blue-door-code-compass. I looked into the two blocking findings on `5e6a858` and believe both are false positives against the current code — details below. A related readability tweak went up in `302bf36`.

**🛑 `waitForJobTerminal` — "`job` out of scope → `ReferenceError`" (`e2e/api-tests/fixtures.ts`)**

`job` is declared with `const` at the top of the `for (;;)` body, and the timeout `throw` is inside that same block, so `job.status` is in scope. No `ReferenceError` is possible, and it already reports the last-seen status:

```ts
for (;;) {
  const job = await getJob(client, jobId);
  if (isTerminalJob(job)) return job;
  if (Date.now() >= deadline) {
    throw new Error(`... last status: ${job.status}`); // in scope
  }
  await new Promise(r => setTimeout(r, JOB_POLL_INTERVAL_MS));
}
```

No change made here — the code is already correct.

**⚠️ `fileForm` — "cast exposes the shared pooled buffer" (`e2e/sdk/client.ts`)**

`new Uint8Array(part.buffer)` already copies into a fresh, exactly-sized buffer, so no pooled bytes were ever uploaded. Confirmed on a pooled Node `Buffer`:

```
Buffer:                     byteOffset=8  len=5  backing.byteLength=8192
after new Uint8Array(buf):  byteOffset=0  byteLength=5  bytes.buffer.byteLength=5
```

The suggested `bytes.buffer.slice(byteOffset, …)` would reduce to `slice(0, 5)` — a redundant extra copy. That said, in `302bf36` I now pass the view straight to `Blob` (`new Blob([bytes], …)`), which is offset-safe by construction and drops the `as ArrayBuffer` assertion — a readability improvement rather than a bug fix.

Happy to reconsider either if I've missed something.

PR #103, opened 2026-09-12
feat(e2e): add design review report with before/after screens

unlikelyzero:
## Summary

- Adds `web/app/e2e/design-review/`: a self-contained HTML before/after of every screen the visual suite covers, built by reading git rather than by running a test. A designer changing a screen regenerates its baseline and commits it, so from that commit on `toHaveScreenshot()` compares the new render against the new baseline and passes — the visual report goes green and empty, and the only record of the previous design is a PNG blob in history that no review tool renders. The change becomes invisible precisely because the workflow was followed correctly. This takes the baseline at the merge base as "before" and the baseline on the branch as "after", so the intended change is what you see.
- **Triggered on a PR by the `design-review` label.** The workflow builds the report, uploads it as the `design-review` artifact, and posts a sticky comment naming the screens that moved and by how much; pushing to a labelled PR refreshes both. Opt-in for the same reason every label gate in this repo is — most PRs do not change a screen. The job runs no Playwright, no backend and no `npm ci`, so it costs seconds. Fork PRs get the artifact but no comment: their token is read-only, and `pull_request_target` would hand a write token to a job running the PR's own code.
- The report inlines its images and computes the comparison in the browser: slider, side by side, onion-skin, and a pixel-difference overlay with adjustable sensitivity. It also lists the templates, styles and assets a branch changed that **no** covered screen renders — a restyled component with no baseline is the one regression a visual suite is structurally unable to report.
- For the comment to state a magnitude, the changed-pixel figure has to exist outside the browser, so `lib/png.mjs` gains a decoder for the 8-bit non-interlaced PNGs a browser writes and `lib/diff.mjs` computes the same comparison in Node. The rule is implemented twice on purpose — Node so the comment and `--json` can quote a number without rendering anything, the browser so the sensitivity slider moves without a round trip — from one shared default threshold, with a test reading that constant out of the browser file to stop the two drifting apart.
- Screens are derived from the baseline filenames, with no registry to keep in sync. The rebase onto the new visual specs took coverage from 6 screens to 11 with no edit to this tool.
- Entry points: `make design-review` / `npm run design:review`, the `design-review` skill, the `design-review` label, and `make design-review-test` wired into `frontend-pr-validation`. The existing visual suite answers the complementary question (did anything change by accident) and is unchanged.

## Test plan

- [x] `make pre-commit`
- [x] Frontend lint/test/build — `make design-review-test` (45 tests), `actionlint` clean on the new workflow, and the full PR check set green including all four e2e shards and the visual project.
- [x] Manually verified the report: generated one for a real merged baseline change (the personality detail redesign) and drove the result in Chromium through all four compare modes, both viewports, both themes, and the sensitivity control, with no console errors or warnings. The Node and browser diffs agree exactly (11.2% desktop / 15.7% mobile). Also exercised the added / removed / unchanged and both missing-merge-base paths.
- [x] Manually verified the label end to end on this PR: the push before labelling skipped, applying the label ran the workflow to success, the artifact downloaded and opened in Chromium with every image decoding, and both a re-run and a subsequent push patched the sticky comment in place rather than posting a second one.

### Review feedback addressed

The rename finding was real, and worse than reported. `changedFiles` parsed git's human-readable output, where a move inside a nested directory is abbreviated to `dir/{old => new}/file`; the regex turned that into `payments}/invoice.component.html` — a path that exists nowhere — and lost the rename status with it, so the file surfaced in the impact and "not covered" lists as something nobody could open. Both commands now use `-z` and are parsed by field, which also fixes paths git backslash-quotes for containing a space or a non-ASCII byte.

Also fixed: the PR lookup validated a ref it never passed to `gh`, while the real hazard was that `gh pr view` resolves the checked-out branch, so any other `--head` would stamp a report about one change with another change's title; a lone `--` (what npm leaves behind) failed the simplest possible invocation; a flag given where a value belongs was swallowed as a ref name, producing a wrong report rather than an error; and `resolveBase` collapsed "you typo'd the ref" and "no common ancestor" into one indistinguishable empty report.

Declined, with reasons: `createImageBitmap` in place of `Image.decode` (needs a blob round-trip from a data URI to solve a problem nothing has measured), and comments instructing readers to update tests when changing a heuristic — the tests already fail in that case, which is the mechanism that works.

## Checklist

### DIFF
--- a/.agents/skills/design-review/SKILL.md
+++ b/.agents/skills/design-review/SKILL.md
@@ -1,0 +1,129 @@
+---
+name: design-review
+description: >-
+  Produce a visual before/after review of a branch or pull request — a
+  self-contained HTML report showing every screen the Playwright visual suite
+  covers, as it looked on main and as it looks now, with a slider/onion/pixel-
+  difference comparison, plus what the change touched and which design files
+  no screen covers. Use whenever someone wants to SEE a change rather than
+  read its diff: "what does my PR look like", "show me the visual impact",
+  "design review for this branch", "did the layout change", "before and after
+  screenshots for the PR", "which screens did I break". Also use when a
+  designer's change lands on a screen with no visual baseline and one should
+  be added.
+---
+
+# Design review
+
+Turns a branch into a picture of itself. The audience is whoever has to
+answer "does this look right?" — usually the person who designed it, who
+should not have to read a unified diff of a Tailwind class list to find out.
+
+## The one thing to understand first
+
+A passing visual suite proves nothing about an intentional design change.
+
+When a designer changes a screen they also regenerate its baseline and commit
+it, so by review time `toHaveScreenshot()` compares the new render against
+the new baseline and passes. The Playwright report is green and empty. The
+change is invisible *because it was done correctly*.
+
+So this report does not read a test run. It reads git: the **before** is the
+baseline committed at the merge base, the **after** is the baseline on this
+branch. That is why it needs no backend, no browser and no Docker, and why it
+finishes in about a second.
+
+The visual suite still matters — it is what catches the *unintended* change,
+where the code moved and the baseline did not. That is a different question
+with its own command; see "When to also run the suite" below.
+
+## Produce the report
+
+From `web/app/`:
+
+```bash
+npm run design:review
+```
+
+Or from the repo root, `make design-review`. Either writes
+`.dev/design-review/report.html` — one self-contained file with the images
+inlined, so it can be sent to someone who will never clone the repo.
+
+Useful variants:
+
+| Goal | Command |
+|---|---|
+| My uncommitted work, against `main` | `npm run design:review` |
+| My branch as committed | `npm run design:review -- --head HEAD` |
+| Someone else's PR | `git fetch origin pull/<N>/head:pr-<N>` then `-- --head pr-<N>` |
+| Everything since a release | `npm run design:review -- --base v1.4.0` |
+| Open it immediately | add `-- --open` |
+| Machine-readable, for a CI assertion | add `-- --json <path>` |
+
+The default `--head` is the **working tree**, not `HEAD`. An uncommitted or
+untracked baseline shows up, which is the point when iterating.
+
+## Read the report back to the user
+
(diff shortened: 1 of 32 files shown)


unlikelyzero:
## What this actually produces

This PR changes no screens, so the automated comment above correctly reads "no screen changed" — which makes it a poor advertisement for the feature. Here is the same tooling run against a real design change instead.

The example is a throwaway restyle of the login page on `demo/design-review-example`: teal pill button instead of an indigo square one, rounder card, larger heading, roomier inputs. Its two baselines were regenerated in the CI container, exactly as a designer would. **The visual suite passes on that change** — the baselines were updated correctly — so nothing else in CI shows what moved. That is the gap this fills.

### The comment it posts

> ### Design review · 1 screen changed
>
> | Screen | Desktop | Mobile |
> | --- | --- | --- |
> | Login page | **5.6%** | **13.3%** |
>
> <sub>10 other screens unchanged.</sub>

Login changed; register stayed identical despite living in the same spec file, as did the other nine screens.

### Drag to compare

The divider follows your pointer. Parked over the button, the old and new sit either side of it:

Comparison slider

### Side by side

Side by side

### Pixel difference

Magenta marks what moved, over a dimmed copy so it stays locatable. Sensitivity is adjustable, and the reading updates live:

Pixel difference

### The whole report

Changed screens first, unchanged below as context, then an impact summary and the list of design files no covered screen renders:

Report overview

### Try it yourself

- **[Get this exact report](https://github.com/theimaginaryfoundation/what-iff/blob/demo/design-review-example/docs/design-review-example/report.html)** → *Download raw file*, then open the saved `.html` in any browser. (Opening the raw URL directly shows the source instead — GitHub serves `.html` as `text/plain`.)
- On a PR, add the **`design-review`** label. The workflow builds the report, uploads it, and the comment links straight at the download. That link needs you signed in to GitHub — artifact downloads always do, even on a public repo.
- Locally, `make design-review`. No backend, no browser, no Docker, no `npm install` — it reads committed baselines out of git and takes about a second.

<sub>`demo/design-review-example` exists only to host this example and is not for merge.</sub>

PR #104, opened 2026-09-12
Add visual coverage for recent UI changes, gated on functional tests

unlikelyzero:
## Summary

- **Adds five visual specs plus baselines** for areas changed over the last three weeks that had no pixel contract: the emoji shortcode autocomplete popup, the empty gallery, the integrations connectors tab, a memory card, and the personality prompt-change diff grid. The committed set goes from 12 PNGs to 22. Two further surfaces get geometric rather than pixel assertions, because their layout has a contract while their content comes from a live turn: the Context X-ray's cost placement and the collapsed prompt-history feed.
- **Makes a screenshot insufficient on its own.** Every file in `tests/visual/` now declares the functional spec behind it, enforced by `check-visual-coverage.mjs` in `frontend-pr-validation`. Applying that rule found two areas standing on nothing: the Context X-ray, whose functional test was disabled everywhere by an unconditional `test.skip(true, ...)` for a flake specific to real-inference backends (now tagged `@mock-only`), and the compaction log, which had no functional spec at all (now added).
- **Fixes a regression the new coverage caught.** Extracting the cost display into `app-context-cost-outlet` moved the styled span one level down from the flex row, so its `margin-left: auto` stopped applying and the estimate sat a few pixels after the budget label instead of trailing at the row's right edge.
- **Checks both viewports on every frontend PR.** The visual step was narrowed by `PROJECTS`, so an unlabelled PR verified only the desktop baselines and the mobile ones waited for a merge to main or the nightly — after a responsive regression had landed.

## Test plan

- [ ] `make pre-commit` (fmt, vet, tidy, test, build)
- [x] Frontend lint/test/build (only if `web/` changed)
- [x] Manually verified: baselines generated in the CI container (`npm run e2e:mock-llm:visual:docker:update`), then confirmed by a separate check-mode run so they reproduce on a second render rather than only matching themselves — 24/24 both times. `npx ng test` 1166 passed. New compaction-log functional spec 2/2. The un-skipped Context X-ray test 5/5 across `--repeat-each=5`. `npm run lint:e2e`, `npm run e2e:check-visual-coverage` and `npm run e2e:check-configs` clean.

## Checklist

- [ ] Updated `docs/ARCHITECTURE_SUMMARY.md` / the affected `_PACKAGE_SUMMARY.md` if this touches system boundaries
- [ ] Regenerated the e2e SDK (`cd web/app && npm run sdk:generate`) if `openapi.yaml` changed

## Notes for reviewers

Two judgement calls worth a second opinion:

1. The cost-alignment fix is **application** code in a PR that is otherwise tests and CI. I found it because the new spec failed, and I nearly weakened the assertion to match the bug before checking the pre-extraction markup. Say so if you would rather it were split out.
2. `check-visual-coverage.mjs` deliberately does not verify that the declared spec covers the *right* behaviour, only that it exists and runs. The intent is to force an author to name something and a reviewer to look at what was named.

Two baselines needed a second pass: a mask sized to a right-aligned timestamp shifts when the rendered time changes width, leaving a sliver of unmasked text. Fixed by removing the variance (screenshot the compaction card's body, mask the memory card's full-width footer row) rather than by raising a pixel tolerance.


unlikelyzero:
Worked through the review. Addressed nine items, declined three, and fixed the one real bug it pointed at. Everything below is in `9db9dc7` and `f474c90`.

### Addressed

**Cost selector could silently drop coverage** — the sharpest one. The geometry checks were gated on `.gauge__cost` matching, so renaming that class, or mounting the private overlay's credit-cost UI here instead, would have skipped the assertions rather than failing them. Now gated on the outlet host rendering any content at all, which is build-agnostic. The text assertion is loosened to a shape so a pricing-precision change can't fail a layout test.

**Save response not asserted** — a 4xx settles `waitForResponse` just as happily as a 200, and the failure would have surfaced as a confusing empty audit list two navigations later. Now asserts `response.ok()`.

**Seeding assumed rather than observed** — `toBeVisible` on the memory card would also have passed if seeding failed and a card with the same fixed content survived an earlier run. Now `toHaveCount(1)`.

**Bounding-box math extracted** — `expectOnSameRow`, `expectTrails`, `expectFlushRight`, `expectVerticallyInside`, plus one `EDGE_TOLERANCE` constant. A failure now names which relationship broke instead of which subtraction produced the wrong number.

**Selectors moved into POMs** — added `contextCostOutlet` on `ChatPage` and `paneWithin()` on `CompactionLogPage` for a card the caller has already narrowed. No spec reaches through a raw class selector now.

**AST memoised by path** — worth doing: several visual specs already cite the same functional spec, so each citation was re-parsing it.

**Failure summary reworded** — it said "without functional coverage" for every problem, including invalid paths and parse failures.

**CSS and spec cross-referenced** — in both directions. `.gauge__top app-context-cost-outlet { margin-left: auto; }` and the right-edge assertion are a pair; changing either alone breaks the other, and that's the point.

**Empty visual directory documented** — the non-zero exit is deliberate. It guards against the check passing because it found nothing, since a path-resolution bug would otherwise look identical to a clean run.

### The race is fixed, not deferred

`JobService.pollJob()` had a genuine ordering bug, and it is the reason the Context X-ray test was skipped in the first place. A job emits several phases that each carry a `result_id`, and each dispatched its own `getMessage()`. Those are independent requests with no ordering guarantee, so the earlier one could resolve last — and because `context_breakdown` is written after the inference phase has already been observed, a late `inference_complete` response carried no breakdown and erased it.

The existing `lastFetchedMessageKey` dedupe does not help: it suppresses a repeat of the *same* `result_id` and phase, and these are different phases.

Each dispatch now takes a sequence number, and a response is applied only if no newer one already has been. Two unit tests cover it — one delivers the phase responses out of order and asserts the breakdown survives, one delivers them in order and asserts the guard drops nothing. The first fails without the fix; I checked by reverting the guard.

With the race gone, the reason for narrowing that e2e test to `@mock-only` no longer holds, so **the tag is off** and it is eligible for the local-LLM runs again — the environment where the race was actually observed.

Note what does and does not prove it. The real-inference workflow is opt-in on PRs and skips by default, so the green "E2E (local LLM backend)" check on earlier commits here ran nothing. I've added the `run-local-model` label to force an actual run on this branch. Until that lands, the evidence for the fix is the pair of unit tests — the out-of-order one fails without the guard — not a real-inference pass.

### Declined

**Centralise persona prompts in a fixture module.** These prompts are assertion inputs, not product copy — the compaction-log spec compares the rendered diff panes against them verbatim. Moving them behind an import means reading two files to understand one assertion, and they only change when a test wants them to.

**Tighten the `@functional-coverage` format.** The parser already handles the variation, and the failure mode is a clear error rather than a silent misparse. Constraining the format would mostly add a rule to remember.

**Factor out the per-spec setup ceremony.** Deliberate. Visual specs earn their keep by being readable top to bottom — the setup is what determines whether a baseline is deterministic, and hiding it behind a helper is how a spec ends up with an unexplained flake. The repetition is the lesser cost.

**AST performance guard.** Memoisation covers the case that exists today, and there is nothing to measure yet. Bailing early on non-Playwright files would be guessing at a cost that has not appeared.

### Verification

Visual suite 24/24 in the CI container after these edits, and 24/24 in CI on both viewports (4.6m). Unit suite 1168 passing, including the two new ordering tests. The X-ray e2e test 5/5 under `--repeat-each` with the tag removed. Analyzer fixtures 11/11. `lint:e2e`, `e2e:check-visual-coverage`, `e2e:check-configs` and `tsc` clean.

PR #119, opened 2026-09-15
feat(data): unified Import & Export screen with selective restore 

GoriVix:
Promotes the account export/import that lived behind `/experimental` into a first-class **`/data` "Import & Export"** screen.

## What's in it
- **`/data` screen** — account **export** (enqueue + poll, email-link delivery) and **restore-from-export**, with **structured** per-section result reporting (personalities / conversations / memories added-vs-skipped, warnings, errors) instead of the old one-line status.
- **Itemized selection ledger on restore** — the chosen ZIP is read **in the browser** (JSZip) to list its personalities and threads; tick which to bring in (select-all/clear), with memories as a single all-or-nothing toggle. Backed by a new optional **`selection`** field on `POST /account/import` (absent = import everything; present = authoritative). `openapi.yaml` + e2e SDK regenerated.
- **Entry points** — an "Import & export your data" button in the settings modal (OSS copy here; the overlay build carries the same in its copy), a "want to import?" link on the no-personalities empty state, and `/experimental` → `/data` redirect. Guard exempts `/data` so fresh users can restore first.
- Removes the dead `experimental` page.

## How to test with prod data
1. Run locally; grab a real account-export ZIP (e.g. the one emailed from prod).
2. Open **Settings → Import & export your data** (or `/data`), or the empty-state link.
3. **Restore** → pick the ZIP → the ledger lists your personalities + threads → select a subset → **Import selected**, and confirm the structured result matches.
4. **Export** enqueues + emails a link (local logs it via NoopSender unless SES is wired).

## Known follow-ups (NOT in this PR)
- ChatGPT/Claude conversation import still lives in the sidebar popup; migrating it onto this screen (and retiring the popup) is the next piece — carries a small product call (sidebar button → `/data`).
- Frontend unit tests for the archive inspector + ledger (deferred; the app builds green).
- The overlay settings-modal entry is a matching change in the private repo (the overlay replaces that modal wholesale).


GoriVix:
Review follow-up:

- Addressed the account-activity concern in 6cf45f4: `ListAccountActivity` now projects the persisted metadata suffix into structured `AccountActivityEntry.metadata`; OpenAPI, the generated E2E schema, and the Angular UI use that field directly. The browser no longer parses audit-log message formatting.
- The summary-index nil-dereference finding is a false positive. `createEmbedding` is a concrete `Handler` method, not a nullable dependency. With the logger-only test fixture it first checks `h.oaiClient == nil` and returns `errMemoryImportUnavailable`; execution continues before `h.ds` is reached. In production `NewHandler` requires the datastore, so adding a nil guard would mask invalid server construction rather than protect a user-facing best-effort path.

csotherden:
A few things worth consideration here:

### P1 — Summary indexing can make large imports fail after data has already been restored

**File:** `internal/handlers/accountexport/import.go`  
**Area:** `importConversationSummaryMemories`

The import now embeds and upserts every exported checkpoint summary serially:

```go
for _, candidate := range summaryImportCandidates(parsed, chatIDs) {
    embeddingVector, err := h.createEmbedding(ctx, candidate.summary)
    ...
    h.ds.UpsertChatSummaryMemory(...)
}
```

This is one embedding request per conversation, unlike the memory importer, which uses bounded batches. On a sufficiently large export, this can consume most or all of the account-import timeout after personalities/conversations have already been written. The job will then be marked failed, leaving a partially restored account and a failure message.

**Recommended repair:**

- Batch summary embeddings, or impose an explicit maximum number of summaries indexed during import.
- Track the phase reached and report a more precise partial-result warning.
- Add a test covering a large number of summaries or embedding timeout/cancellation.
- Consider indexing summaries asynchronously after the primary restore reaches `complete`, if search availability is genuinely best-effort.

### P2 — `selection` has no server-side size or cardinality limits

**File:** `internal/handlers/accountexport/import.go`

The multipart `selection` field is parsed before staging:

```go
if raw := strings.TrimSpace(r.FormValue("selection")); raw != "" {
    var sel models.AccountImportSelection
    if err := json.Unmarshal([]byte(raw), &sel); err != nil {
        ...
    }
    selection = &sel
}
```

There is no bound on:

- JSON byte length
- `personality_ids` length
- `conversation_ids` length
- duplicate IDs

An authenticated caller can submit a very large selection payload and force allocation, UUID parsing, map construction, and filtering before the archive is even validated. This is primarily a resource-exhaustion issue rather than an authorization bypass.

**Recommended repair:**

- Enforce a maximum selection payload size.
- Enforce maximum IDs per array.
- Optionally reject duplicates or normalize them explicitly.
- Add handler-level tests for oversized and malformed-but-valid selections.

### P2 — Browser-side archive inspection has no resource guard

**File:** `web/app/src/app/features/data-portability/account-archive.service.ts`

`JSZip.loadAsync(file)` is called before any client-side size or entry-count limit, and `conversations.json` plus every personality entry are then read in the browser. The server has archive safety limits, but those do not protect the browser from a large or malicious ZIP selected locally.

This can freeze or exhaust the tab before the upload reaches the server.

**Recommended repair:**

- Reject files above a reasonable byte limit before `loadAsync`.
- Check archive entry count and uncompressed-size metadata where available.
- Avoid reading every personality/conversation entry until the archive passes basic bounds.
- Keep the server-side validation authoritative; this is only a client UX/resource guard.

PR #120, opened 2026-09-15
Provider API keys belong to accounts, not the process

LiminalDevving:
> **Stacked PR — 1 of 8.** Base is `main`. Merge first; the other seven build on it.

## Summary

- Model-provider API keys move from process-level environment variables to **per-account** storage, so two people on one deployment use their own keys and their own billing. Keys are encrypted at rest, write-only over the API, and surfaced only as a four-character hint.
- A resolver caches an account's key and is invalidated on write, so a new key takes effect on the next request rather than at the next restart.
- The model list only offers models whose provider has a usable credential, instead of offering everything and failing at send time.
- `make init-env` generates a working `.env` from a fresh clone, so first-run setup is one command rather than a checklist.

Deployment-level environment keys still work and act as the fallback when an
account has supplied nothing.

## Test plan

- [x] Manually verified: stored a key through the API, confirmed it is returned only as a hint, confirmed an account key overrides the deployment fallback and that removing it falls back again

## Checklist


LiminalDevving:
Putting this back to draft.

ent/schema/user.go
```diff
@@ -88,6 +88,10 @@ func (User) Edges() []ent.Edge {
 			Annotations(entsql.OnDelete(entsql.Cascade)),
 		edge.To("mcp_servers", MCPServer.Type).
 			Annotations(entsql.OnDelete(entsql.Cascade)),
+		// Cascade so deleting an account takes its provider credentials with
+		// it, the same as every other per-user secret here.
+		edge.To("provider_keys", UserProviderKey.Type).
```
GoriVix, line 93: this should use the extensions hook in OSS as we do not want this variable in the deployed system

internal/agent/provider/openai_credential.go
```diff
@@ -0,0 +1,86 @@
...
+	k := strings.TrimSpace(key)
+	return func(context.Context) string { return k }
+}
+
+type openAICredentialTransport struct {
+	resolve OpenAIKeyResolver
+	base    http.RoundTripper
+}
+
+func (t *openAICredentialTransport) RoundTrip(req *http.Request) (*http.Response, error) {
```
GoriVix, line 49: what is 'round trip'? It is not clear what this class is doing / supposed to do

internal/models/provider_availability.go
```diff
@@ -0,0 +1,128 @@
...
+	keyed map[ModelProvider]bool
+	// filter is false when provider credentials do not determine which models
+	// work, in which case every model is reported available.
+	filter bool
+	// openAI, when set, is consulted instead of the static entry. OpenAI keys
+	// belong to accounts and can be set at runtime, so availability is a
+	// question about the caller rather than about the process: a user who has
+	// just entered a key expects their models to appear, and a user who has
+	// not should not be offered models that will fail for them.
+	openAI func(context.Context) bool
```
GoriVix, line 24: why is this called openAI? not sure what this is meant to do

internal/providerkeys/resolver.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/google/uuid"
+	"github.com/theimaginaryfoundation/what-iff/internal/apicontext"
+	"github.com/theimaginaryfoundation/what-iff/internal/datastore"
+)
+
+// cacheTTL bounds how long a key change takes to reach in-flight work that
+// resolved before the change. Writes invalidate the entry immediately, so this
+// only covers the case of another process updating the same row.
+const cacheTTL = 5 * time.Minute
```
GoriVix, line 23: this is not a real use-case for OSS. You cant really have two processes running locally (or there is no reason to do so) and we do not use this in the deployed version since keys are managed differently

internal/server/server.go
```diff
@@ -133,6 +141,16 @@ func (s *Server) setupRoutes() {
...
+	} else {
+		// Vendor path: route OpenAI-family clients through a transport that
+		// resolves the key per request from the account making it. The SDK
+		// client value is copied into six independent holders, so there is no
+		// single field to reassign; resolving at the transport they all share
+		// reaches every one without touching a call site. Scoped to the OpenAI
+		// host so the other OpenAI-compatible providers keep their own
+		// credentials (see openai_credential.go).
+		s.openAIKeys = providerkeys.NewResolver(dataStore, string(appmodels.ModelProviderOpenAI), s.config.OpenAIKey)
+		providerHTTPClient = provider.OpenAICredentialHTTPClient(s.openAIKeys.Resolve, nil)
```
GoriVix, line 153: this comment is confusing to me + the code seems to not do anything. Why do we need a resolver for the static key case? 

It would make more sense to create an interface for resolving the keys and then have the client interfaces take in *that* as an input variable and have the default be 'static key resolver' and then for OSS we can replace that with the dynamic key resolver.

GoriVix (changes requested):
Some comments. This PR also does not seem very self-contained, I feel like I probably need to read some of the later PRs to get an understanding of what is going on here. 

In the future, that is a smell -- if you need multiple PRs you should write a ADR outlining the process and each PR should be deployable in isolation

PR #121, opened 2026-09-15
Per-account credentials for every provider, not just OpenAI

LiminalDevving:
> **Stacked PR — 2 of 8.** Base is the previous PR's branch (#120), not `main`, so the
> diff shown is only this PR's own changes. Merge after #120.

## Summary

- Every provider client is constructed unconditionally. A client was previously built only when a process-level key existed at startup, and one built without a key could not be given a key later — which is why only OpenAI supported per-account credentials.
- Credentials are applied **per request** through an HTTP transport scoped by host, so the same client serves different accounts with different keys.
- Nine nil-guards come out. A provider client is always present, so "is this provider available" becomes a question about the account's credential rather than about whether an object exists.
- A missing credential now produces an error naming the provider and the model, instead of a nil dereference.

Error text stays deployment-neutral: it states what is missing, not where to
fix it, because which remedy applies depends on how the deployment is run. The environment variable name goes to the logs, not to the response.

## Test plan

- [x] Manually verified: a chat on a provider with no configured key returns a clear error rather than panicking, and the same build serves two accounts with different keys

## Checklist


LiminalDevving:
Putting this back to draft.

internal/agent/provider/claude.go
```diff
@@ -320,6 +320,9 @@ func (c *ClaudeProvider) betaMessagesNewStreaming(
 // Call sends a single Messages API request. Retry handling (429/500/529) is
 // delegated to the Anthropic SDK via WithMaxRetries in NewClaudeProvider.
 func (c *ClaudeProvider) Call(ctx context.Context, params anthropic.MessageNewParams) (*anthropic.Message, error) {
+	if c == nil || c.client == nil {
+		return nil, ErrProviderUnavailable
+	}
```
GoriVix, line 325: ideally we should not use nil as a valid 'unconfigured' state, risky for system crashes.

Also, c cant be nil we are within c (I think? you are making me question golang things now)

internal/agent/provider/fileattachment.go
```diff
@@ -71,6 +71,9 @@ func normalizeUploadFileNameExtension(fileName string) string {
 }
 
 func (a *OpenAIProvider) DeleteFileAttachment(ctx context.Context, fileID string) error {
```
GoriVix, line 73: do we still even use this? I dont think we ever upload / download from O.AI file store anymore? (maybe we still upload for openai chats?) If not we should just clean this up

internal/agent/provider/openai.go
```diff
@@ -184,6 +187,9 @@ func waitForRetry(ctx context.Context, d time.Duration) error {
 }
 
 func (c *OpenAIProvider) CountTokens(text string) (int, error) {
+	if c == nil {
+		return 0, ErrProviderUnavailable
+	}
```
GoriVix, line 192: c cant be nil here because we are inside it already

internal/agent/provider/openai_credential.go
```diff
@@ -33,54 +20,3 @@ const openAIAPIHost = "api.openai.com"
...
-
-// StaticOpenAIKey resolves to the same key for every actor. Used for the
-// deployment-wide fallback and in tests.
-func StaticOpenAIKey(key string) OpenAIKeyResolver {
-	k := strings.TrimSpace(key)
-	return func(context.Context) string { return k }
-}
-
-type openAICredentialTransport struct {
-	resolve OpenAIKeyResolver
```
GoriVix, line 45: this appears to be deleting the code you submitted in the PR upstack, which is a smell. Did it move somewhere? What happened?

These sorts of things are stuff you should do manual PR review for and fix pre-review :)

internal/agent/message.go
```diff
@@ -133,6 +134,10 @@ type Agent struct {
 	// is chosen at construction and swapped via server wiring, so no push detail
 	// reaches this package.
 	pushNotifier pushnotify.Notifier
+	// providerKeys answers, per request, which providers the calling account can
+	// use. Nil under mock/local backends and in tests, where credentials are not
+	// consulted at all.
+	providerKeys *providerkeys.Registry
```
GoriVix, line 140: this name mixes concerns, it *says* it is a provider *Registry* but then you call it keys, which implies it is an api-key retrieval interface

internal/agent/message.go
```diff
@@ -299,50 +309,48 @@ func NewAgent(ds *datastore.Datastore, logger *zap.Logger, tel *telemetry.Teleme
...
+	// same as "unavailable" any more — an account supplying its own key makes
+	// the provider usable for that account — so it is phrased as what it is.
+	for name, key := range map[string]string{
+		"ANTHROPIC_API_KEY": anthropicKey,
+		"ZAI_API_KEY":       cfg.ZAIKey,
+		"GEMINI_API_KEY":    cfg.GeminiKey,
+		"MISTRAL_API_KEY":   cfg.MistralKey,
+		"DEEPSEEK_API_KEY":  cfg.DeepSeekKey,
+		"QWEN_API_KEY":      cfg.QwenKey,
+		"XIAOMI_API_KEY":    cfg.XiaomiKey,
```
GoriVix, line 348: I am kinda worried we are making this far too complex here. I am not sure "I have two local accounts with different API keys" is even a valid use case, and now we TWO layers -- the env default and the user configured. I'm not sure the .env values are useful at all now for OSS. We probably want a keyResolver() and in deployed it can pull from env, for local it *only* uses whatever is on the user account, that would match other seams we have made and reduce the overlapping use-case complexity

internal/agent/provider_keys.go
```diff
@@ -0,0 +1,67 @@
...
+var providerLabels = map[models.ModelProvider]struct{ name, envVar string }{
+	models.ModelProviderOpenAI:    {"OpenAI", "OPENAI_API_KEY"},
+	models.ModelProviderAnthropic: {"Anthropic", "ANTHROPIC_API_KEY"},
+	models.ModelProviderZAI:       {"z.ai", "ZAI_API_KEY"},
+	models.ModelProviderGoogle:    {"Gemini", "GEMINI_API_KEY"},
+	models.ModelProviderMistral:   {"Mistral", "MISTRAL_API_KEY"},
+	models.ModelProviderDeepSeek:  {"DeepSeek", "DEEPSEEK_API_KEY"},
+	models.ModelProviderQwen:      {"Qwen", "QWEN_API_KEY"},
+	models.ModelProviderXiaomi:    {"Xiaomi", "XIAOMI_API_KEY"},
+}
```
GoriVix, line 29: if this survives, we should NOT inline the strings here for the env variables but instead pull constants from a spot that is used both here AND where actually referenced at startup. That way a typo or variable change wont lead to these being out of sync

GoriVix (changes requested):

PR #122, opened 2026-09-15
Add a provider API key from the app instead of editing .env

LiminalDevving:
> **Stacked PR — 3 of 8.** Base is the previous PR's branch (#121), not `main`, so the
> diff shown is only this PR's own changes. Merge after #121.

## Summary

- Adds an **API Keys** tab under Integrations, so a key is added in the app rather than by editing `.env` and restarting.
- Each provider row shows whether it is configured and where the credential comes from — the account's own key or the deployment's — so "it works" and "it works because of someone else's key" are distinguishable.
- Every provider gets a "Where do I get one?" link, because a setup screen that assumes you already know is not a setup screen.
- A setup guard sends an account with no usable key here rather than into a chat that cannot answer.

## Test plan

- [x] Manually verified in a browser: added a key, replaced it, removed it, and confirmed the row reports the right source each time; confirmed the guard redirects and then releases

## Checklist


LiminalDevving:
Putting this back to draft.

GoriVix (approved):
LGTM two questions

1. can you upload screenshots of the new screens and make sure they are covered in the screen / integ tests
2. Just verifying that this is an integration seam and we can turn it off in the SaaS deploy

PR #123, opened 2026-09-15
Choose what the model picker shows you

LiminalDevving:
> **Stacked PR — 4 of 8.** Base is the previous PR's branch (#122), not `main`, so the
> diff shown is only this PR's own changes. Merge after #122.

## Summary

- An account can **hide models from its own picker**. The instance-level catalog stays shared; what you scroll past is yours. A hidden model keeps working in chats and jobs that already use it — this changes the picker, not the data.
- Adds a **Models** tab beside API Keys, grouped by provider to match how the picker presents them.
- Subscription tiers move behind a seam. Tier is a billing concept, and a build with no billing has no use for a Tier tab or tier badges. The picker already defaulted to grouping by vendor, so nothing changes for anyone who was not using that tab.
- Fixes an off-by-one in the "keep at least one model visible" rule: the warning appeared when you tried to hide your last model but only cleared once you had **two**, so it sat there for as long as you stayed at one. The constraint is now a disabled control and a quiet note rather than an error raised after the click.

## Test plan

- [x] Manually verified in a browser: hid and restored models, confirmed the picker follows, confirmed a hidden model still works in a chat that already used it

## Checklist


LiminalDevving:
Putting this back to draft.

GoriVix:
Same as other one -- UI screenshots. You have a compass issue still as well.

What motivates doing this one right now? I am fine with it and we have talked about having something like this for the live service too (although I was considering removing them from the listModels() call as well, so agents dont create *new* jobs or subagents with disabled models -- that way the feature also can be 'dont spin up subagents with fable plz k thx') and are we planning to release this one over in the SaaS side and does it work for that?

PR #124, opened 2026-09-15
Assumptions a self-hosted install cannot meet

LiminalDevving:
> **Stacked PR — 5 of 8.** Base is the previous PR's branch (#123), not `main`, so the
> diff shown is only this PR's own changes. Merge after #123.

## Summary

Four places where the app assumed capabilities an ordinary account does not have.

- **Service tiers.** Requests hardcoded `priority` and `flex`, which an ordinary account cannot use. Now omitted unless `OPENAI_SERVICE_TIERS=true`.
- **The first-chat greeting** was pinned to an Anthropic model. It fires automatically on account setup, before the user has configured anything, so an install without an Anthropic key opened on a missing-key error. Moved to an OpenAI model — the one provider chat already requires.
- **`gpt-5.1-nano` does not exist.** Seeded from the open-source release onward; a request returns 404 `model_not_found`, with none of the aliasing to a dated snapshot that real names get. Anyone who picked it had every send fail. Removing it from the seed catalog only helps new installs, so it is retired through the existing migration, which also repoints affected chats and account defaults.
- **Archival required Anthropic regardless of the chat's provider.** Gemini and GLM chats take the same rebuilt-context checkpoint path as Claude chats, and its scratchpad and memory steps were hardcoded to Anthropic. Those chats worked while memory silently never updated. Archival now follows the chat's own provider.

Also swaps the shared small-utility model, whose current pin has a shutdown
date about five weeks out, and consolidates it into one constant — four jobs
across two packages shared that string by copy.

## Test plan

- [x] Manually verified: every model the app depends on checked against the provider's live model list; retirement verified against a real database — first boot migrated and soft-deleted the row, the model list dropped by one, and a second boot was a clean no-op

## Checklist


LiminalDevving:
Putting this back to draft.

GoriVix:
> Service tiers. Requests hardcoded priority and flex, which an ordinary account cannot use. Now omitted unless OPENAI_SERVICE_TIERS=true.

Are you sure about this? I was under the impression that the tiers were available to anyone. OpenAI is (afaik) the only vendor that has these in general though. I'm fine with making it a local toggle

internal/agent/archival_from_context.go
```diff
@@ -0,0 +1,166 @@
...
+// The checkpoint sequence has two shapes. OpenAI chats chain each archival step
+// off the response id of the turn that just finished. Every other provider
+// rebuilds conversation context from the database and sends it explicitly.
+//
+// That second shape used to be hardcoded to the Anthropic provider, which was
+// correct only by accident: it is also the path Gemini and z.ai chats take, and
+// those accounts have no reason to hold an Anthropic key. The result was a chat
+// that worked while its memory silently never updated. These functions are the
+// OpenAI rendering of the same steps, for chats whose own provider is not
+// Anthropic — OpenAI being the one credential every account already needs.
```
GoriVix, line 29: Ok. So we have what, 8 model providers. There should be a path to go to OpenAI that is NOT OpenAI native, so doesnt start with a previousResponseId. Those paths DO generally continue to use previousResponseId for subsequent calls in the summarizer chain, which they dont really need to but it is a legacy / convenience.

If we are going to change this, we should make *every* model go through the exact same summarizer and assume no chaining of responseIds *into* the summarizer (they can chain *within* the summarizer mostly not worth tearing this down)

internal/agent/archival_from_context.go
```diff
@@ -0,0 +1,166 @@
...
+//
+// archivalContextItems renders the rebuilt context, then appends the step's own
+// instruction turns. Images the Responses API will not accept are stripped
+// first: one unsupported attachment otherwise fails the whole call with a 400,
+// which is the same guard the checkpoint summariser already applies.
+func archivalContextItems(modelContext *provider.ModelContext, extra ...responses.ResponseInputItemUnionParam) []responses.ResponseInputItemUnionParam {
+	items := provider.RenderOpenAIInputItems(provider.SanitizeImagesForOpenAIInput(modelContext))
+	return append(items, extra...)
+}
```
GoriVix, line 38: this should almost certainly exist on the modelContext object, if anywhere. There is already a method to add items / clone

internal/agent/first_chat_onboarding_prompt.go
```diff
@@ -7,7 +7,19 @@ import (
...
+// It must name an OpenAI model. The greeting fires automatically on account
+// setup, before the user has chosen anything, so it cannot depend on a
+// provider the account may not have configured — pointing it at Anthropic
+// made the first thing a new self-hosted user saw an "Anthropic API key is
+// not configured" error. OpenAI is the one provider chat already requires,
+// so an OpenAI model is the only choice that is always satisfiable.
+//
+// This must also be a model seeded in models.AvailableModels, because the
+// greeting is resolved through the model catalog by name.
+const FirstChatGreetingModelName = "gpt-5-mini"
```
GoriVix, line 22: we should standardize this to the same variable as the other defaults and use 5.6-luna

GoriVix (commented):
We should split this one up -- it mixes very low risk things (The tiering, changing models for first message) with VERY high risk things (fucking with the summarizer logic). I'm gonna need to go through the summarizer changes with a fine-tooth comb, bugs there are VERY subtle and hard to notice and the code is not straightforward, it is the main place in our app where there be dragons :(

PR #125, opened 2026-09-15
What each provider key is for, and what it can reach

LiminalDevving:
> **Stacked PR — 6 of 8.** Base is the previous PR's branch (#124), not `main`, so the
> diff shown is only this PR's own changes. Merge after #124.

## Summary

- Adds a **providers page** answering "what is my key being spent on, and by which model", replacing a link to a README anchor that did not exist. It is read mid-setup by someone deciding which keys to paste, so it belongs in the app.
- The job-to-model mapping is served from `/provider-usage` and built from the same constants the calls use, rather than written out in the page. A second copy is how the architecture summary came to name the wrong archival model for months.
- Adds `/provider-models/{provider}`, which asks a provider which models it actually serves, using the calling account's key. The hand-written catalog rotted; a list sourced from the provider cannot contain a name the provider does not have, and because it uses the caller's own key, what comes back is what that key can reach.

Nothing is filtered server-side. Retired and non-chat entries are returned with
their kind and shutdown date, because the kind is a heuristic over the model id
— these list endpoints return no capability information at all — and dropping a
misjudged model would make it unreachable rather than merely mislabelled.

Only OpenAI has a lister; other providers answer 501 rather than pretending.

## Test plan

- [x] Manually verified against a live provider: full catalog returned and bucketed, retired entries flagged, cached reads served without a round trip, `?refresh=true` bypassing the cache, and an unimplemented provider answering 501

## Checklist


LiminalDevving:
Putting this back to draft.

GoriVix (commented):
This one should also be split up into the easy stuff (changing base models, setting constants) and the new functionality (new UI screen)

Also please add UI screenshots + test coverage.

Also we dont need caches on this stuff I think, it is extra complexity without much value -- this is like 1kb of DB data at most and the cache TTL stuff can create confusing interactions to users / more seams to maintain.

PR #128, opened 2026-09-16
Read an unset chat rehydration state as empty rather than failing

LiminalDevving:
Standalone — not part of the `oss/*` stack, and independent of it.

## What rehydration is

Imported threads arrive as a raw transcript with none of the context a native chat builds as it goes — no checkpoint summary, no extracted memories. Rehydration is the one-time background job that reconstitutes that: it summarises the transcript and keeps the last few turns verbatim so the next reply is continuous rather than obviously reading notes. It is expensive on long threads, so it runs lazily, when the thread is unarchived.

`WaitForThreadRehydration` is the gate in front of it. Rehydration runs in the background, so a user can restore a thread and send a message while the summary is still being written; the gate stalls that turn until rehydration reaches a terminal state, or proceeds without it after a timeout. For a thread that needs no rehydration — every natively created chat — it returns immediately.

## Summary

`rehydration_state` is optional and left unset on natively created chats, so the value in the column is normally NULL. The gate scanned it straight into a string, which cannot convert NULL, so it errored and logged on every message in every native chat.

It fails open, so the turn proceeded and nothing broke. Two things were wrong anyway. The gate could not do its job for any chat in that state — had one genuinely had rehydration in flight, it would have errored rather than waited, and the reply would have gone out without the summary. And a warning on every message makes a real rehydration failure indistinguishable from the normal background hum.

Reads the row rather than the column. Ent maps an unset optional string to `""`, which is already the documented value for "no rehydration needed".

## Test plan

- [x] Unit tests cover the unset and populated cases
- [x] Manually verified against a live database: the same chat with a NULL state logged a failure on every message before this change and none after

## Checklist


GoriVix (approved):
RIP good catch

PR #131, opened 2026-09-20
fix: Gemini token double-count, OpenAI forced-final threading, truncation error

GoriVix:
## Summary

Three provider-turn bugs that each ate a whole chat turn:

- **Gemini token inflation.** `streamChatCompletion` relied on `openai.ChatCompletionAccumulator`, which sums `chunk.Usage` across every SSE chunk. That is correct for OpenAI (usage only in the final chunk) but wrong for Gemini's OpenAI-compatible streaming, which repeats a full **cumulative** usage block on every chunk — so prompt tokens were multiplied by the chunk count (an 80k prompt reported as ~300k once the reply got long). Now the usage from the last chunk that carries any is used as the authoritative total. Fixes the inflated per-turn token metric for all OpenAI-compatible providers (Gemini, Mistral, DeepSeek, Qwen, Xiaomi).
- **OpenAI `400 No tool output found for function call`.** `Call` sets `previous_response_id` at request entry but advances the stored id only *after* the call returns, so `params.PreviousResponseID` lags one round behind. The normal loop re-syncs it each `Call`; `ForceFinalResponse` did not. When the model requested tools on both the next-to-last and last loops, the forced final call paired the last round's `function_call_output` items against the *previous* round's response id and OpenAI rejected the turn. `ForceFinalResponse` now re-threads `previous_response_id` to the latest response before firing.
- **Opaque truncation failure.** When a turn spends its whole `MaxTokens` (8192) budget on reasoning or a partial tool call and is cut off before any text block (seen on tool-heavy GLM turns), the empty-turn guard correctly refuses to persist a blank message but surfaced a raw internal dump as the user-facing error. It now detects output-length truncation (`max_tokens` / `max_output_tokens`) and returns a clear, actionable message; the turn still fails and stays retryable. Note: this makes the *failure* legible — the underlying question of why GLM spends its budget on reasoning tokens is separate follow-up work.

Each fix has a regression test that fails without the change:
`TestStreamChatCompletion_DoesNotSumRepeatedUsage`,
`TestOpenAIAdapter_ForceFinalResponse_ThreadsLatestResponseID`, and the
truncation cases in `TestAssertGenerationProducedOutput`.

## Test plan

- [x] `make pre-commit`
- [ ] Frontend lint/test/build
- [x] Manually verified: Gemini token metric holds steady across turns instead of ballooning, and the OpenAI forced-final tool-loop no longer 400s (both reproduced and confirmed fixed against a local Docker stack)

## Checklist


GoriVix:
Compass blocker is a false positive, that is not changed in this PR

PR #139, opened 2026-09-21
feat(chat): remove bookmarks from the navigator and speed up jump-back

### LINKED ISSUE #137: Improve bookmark navigation performance and loading feedback in long threads
### Problem

Selecting a bookmark in a long thread can require loading many pages and scrolling through a large amount of history before the bookmarked interaction is reached. The UI may appear stalled or freeze during that work, and it does not always clearly communicate that the target is being located or loaded.

### Desired behavior

When the bookmark target is known, load or locate the relevant portion of the thread efficiently and navigate to the bookmarked interaction with a visible loading state while keeping the interface responsive.

The implementation may increase the page/batch size, request most or all of the thread once the bookmark signal is available, use targeted range/offset loading, or combine these approaches with virtualization. Choose based on measured latency, memory, and responsiveness tradeoffs.

### Acceptance criteria

- Selecting a bookmark does not require the user to manually scroll through the entire thread to reach the target.
- The target interaction is brought into view and visibly indicated/highlighted once available.
- A visible loading state communicates that the bookmarked interaction is being located or loaded.
- Loading feedback is cleared when navigation succeeds, fails, is cancelled, or the target is unavailable.
- The interface remains responsive during loading; no unbounded synchronous work freezes the UI.
- The behavior works for bookmarks near the beginning, middle, and end of long threads.
- List loading, thread loading, and target-location failures have distinct, recoverable error states.
- Existing thread state is preserved where practical when navigation is cancelled or fails.
- Add regression coverage for long-thread bookmark navigation, slow-network behavior, and loading/error states.
- Instrumentation or request logging makes page size, request count, and target-location latency observable.

GoriVix:
Closes #137.

Two bookmark-navigation quirks in long threads.

## 1. Un-bookmark from the navigator dropdown
Previously you had to scroll back to a message to un-star it. Each row in the bookmark popup now has a trailing gold-star toggle that removes the bookmark via the existing `setBookmark(false)` path. The menu **stays open** so several can be cleared in a row, and `stopPropagation` keeps the remove click from also triggering a jump.

## 2. Snappy jump-to-bookmark (the core of the issue)
Jumping to a bookmark far back in a thread walked older history one **50-message offset page at a time** — dozens of sequential roundtrips.

The root cause is that the descending message view used offset pagination (`offset = (page-1) × pageSize`), which couples batch size to a fixed page size, so you can't just request a bigger batch. This replaces the "older" direction with **keyset pagination**:

- Backend `GET /chat/{id}/chat-message` accepts an opaque `cursor` (URL-safe base64 of `sent_at|id`); when present it keysets `(sent_at, id) < cursor` newest-first instead of doing offset math. Responses carry a `next_cursor` to continue the walk. **Additive** — `page`/offset still works; initial load and live reconcile are untouched.
- The cursor codec lives in `models` (`EncodeMessageCursor`/`DecodeMessageCursor`) as the shared contract between the datastore and the handler.
- Frontend scroll-back and jump-to-bookmark now ride the cursor. The jump uses a **200-message batch**, so a far-back bookmark resolves in a handful of roundtrips instead of dozens. The dead `messagesPage` counter is gone.
- A header status line shows *"Locating bookmark…"* and clears on arrival (the existing flash is the success cue) or reports *"Could not locate that bookmark."* — covering the issue's loading/failure-feedback criteria.

## Tests & tooling
- Go: `TestListChatMessagesBefore_KeysetWalk` (multi-batch walk, varying batch sizes, `total_count` semantics, empty-tail cursor) + cursor codec round-trip/garbage-rejection tests.
- Frontend: message-service cursor/prepend behaviour, session-service cursor walk (reaches target / stops when cursor runs out), and a new navigator-component spec (remove emits without jumping and keeps the menu open). Full suite: 1425 passing.
- Regenerated the e2e SDK from `openapi.yaml`; updated the datastore `_PACKAGE_SUMMARY.md`.

## Follow-up (not in this PR)
An end-to-end Playwright test for a long-thread jump would be a nice regression guard but needs the mock-LLM stack; happy to add it separately if wanted.


GoriVix:
Compass stuck on unrelated diffs / changes again, not relevant to this PR

